PR #9, opened 2017-08-18
Add TokenLocation

pynicolas:


src/main/java/org/sonarsource/analyzer/commons/TokenLocation.java
```diff
@@ -0,0 +1,58 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.analyzer.commons;
+
+public class TokenLocation {
```
vilchik-elena: @pynicolas doc?:)

pynicolas:
@vilchik-elena Added javadoc in 1361238

nixel2007:
Wanna share out implementation of TokenLocation with handling the BOM.

```java
  public TokenLocation(Token token, boolean hasBom) {
    int columnCorrection;

    this.startLine = token.getLine();
    // WARNING:
    // If there is BOM in the file (even it was excluded at lexer chanel)
    // first token has column number with offset +1
    // To prefent crash of highlighter and scanner, column correction is needed.
    // If you remove this correction, tests WILL PASS, but the sonar-scanner fails
    if (hasBom && startLine == 1) {
      columnCorrection = -1;
    } else {
      columnCorrection = 0;
    }
    this.startLineOffset = token.getColumn() + columnCorrection;

    String value = token.getValue();
    String[] lines = value.split("\r\n|\n|\r", -1);

    if (lines.length > 1) {
      endLine = token.getLine() + lines.length - 1;
      endLineOffset = lines[lines.length - 1].length();

    } else {
      this.endLine = this.startLine;
      this.endLineOffset = this.startLineOffset + token.getValue().length();
    }
  }
```

Note sure if 6.4 fix this behavior with https://jira.sonarsource.com/browse/SONAR-9204 and https://jira.sonarsource.com/browse/SONAR-9199.

nixel2007:
direct ping to @vilchik-elena and @pynicolas (not sure if you'll recieve a mail notification)

pynicolas:
Thanks for sharing @nixel2007 !
I think I prefer to keep the handling of BOMs outside of this class.
The tokens which we store in the AST should already have a correct line and column.

nixel2007:
> The tokens which we store in the AST should already have a correct line and column.

Agree. But I couldn't do this "easily" at parser/lexer level year ago (5.6 or 6.0 API), so this implementation was born. I think, I need to check lexer output at 6.4 API)

PR #21, opened 2017-09-26
Add comment parser

### LINKED ISSUE #20: [test-commons] Some plugins can't use the AST to extract comment.
SonarPLSQL don't use the AST to extract comments because of the parse errors. But it still wants to test the result.
SonarCobol can't use the AST to extract comments because to comments are truncated at column 73. And SQL comment are not present.
The code to parse comment could be factorized in the test-commons library.

alban-auzeill:
Fixes #20


test-commons/src/main/java/com/sonarsource/checks/verifier/CommentParser.java
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * <pre>
+   * This comment parser is able split a line using "commentPrefix".
+   * But there's some limitation, like for this case in java:
+   * String name = "Paul//Smith"; // Noncompliant
+   * Example:
+   *   parser.addSingleLineCommentParser("//");
+   * </pre>
+   */
+  CommentParser addSingleLineCommentParser(String commentPrefix);
```
pynicolas: I suggest renaming this method to "addSingleLineCommentSyntax".

test-commons/src/main/java/com/sonarsource/checks/verifier/CommentParser.java
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * <pre>
+   * This comment parser find comment delimited by "commentPrefix" and "commentSuffix".
+   * But there's some limitation, like for this case in java:
+   * String name = "/*";
+   * Example:
+   *   parser.addMultiLineCommentParser("{*", "*}");
+   * </pre>
+   */
+  CommentParser addMultiLineCommentParser(String commentPrefix, String commentSuffix);
```
pynicolas: Let's remove that if we don't really need it.

PR #25, opened 2017-09-27
Support several messages on the same line regardless of the occurrence order.

### LINKED ISSUE #24: Message order on the same line should not matter
The unit test of the Cobol rule ForbiddenKeywordInCopybookCheck is able to generate 21 issue on the same line. Providing all the issue messages in the wrong order should me supported.

alban-auzeill:
Fixes #24


pynicolas:
Looks good to me.

PR #31, opened 2018-01-03
Fix IllegalStateException with no message when line number > 999, fixes #30

### LINKED ISSUE #30: IllegalStateException with no message when line number > 999
CheckVerifier should not fail with an exception with no message when a tested file has more than 999 lines. CheckVerifier should support test files with any number of lines. See https://github.com/SonarSource/sonar-analyzer-commons/blob/8b4605b4f82ca2e85e2ec701b507eefd1f953216/test-commons/src/main/java/com/sonarsource/checks/verifier/internal/LineIssues.java#L162

alban-auzeill:


travis.sh
```diff
@@ -4,10 +4,11 @@ set -euo pipefail
 
 function configureTravis {
   mkdir -p ~/.local
-  curl -sSL https://github.com/SonarSource/travis-utils/tarball/v38 | tar zx --strip-components 1 -C ~/.local
+  curl -sSL https://github.com/SonarSource/travis-utils/tarball/v43 | tar zx --strip-components 1 -C ~/.local
   source ~/.local/bin/install
 }
 configureTravis
+source ~/.local/bin/installMaven35
```
vilchik-elena: @alban-auzeill what is this for?

alban-auzeill: There's two commits, this change is part of "Update travis-utils from v38 to v43"
travis-utils v38 was not working, and v43 need "installMaven35" to not raise the following error:
```
[ERROR] No plugin found for prefix 'sonarsource' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories
```

PR #40, opened 2018-02-05
Add recognizers from sslr-squid-bridge

### LINKED ISSUE #38: Move recognizers from sslr-squid-bridge into dedicated module

m-g-sonar:
fixes #38

### DIFF
--- a/recognizers/pom.xml
+++ b/recognizers/pom.xml
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.sonarsource.analyzer-commons</groupId>
+    <artifactId>sonar-analyzer-commons-parent</artifactId>
+    <version>1.5-SNAPSHOT</version>
+  </parent>
+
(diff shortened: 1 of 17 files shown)


recognizers/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+  <description>Shared code recognizers</description>
+
+  <dependencies>
+    <!-- COMPILE -->
+    <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>commons-lang</groupId>
```
m-g-sonar: I'm not sure about this dependency (especially regarding its scope).

pynicolas: I think we need it because of [ContainsDetector](https://github.com/SonarSource/sonar-analyzer-commons/pull/40/files#diff-f038011876aeb5159363030078e05429R24) and with a "compile" scope. This declaration looks ok.

recognizers/pom.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.sonarsource.analyzer-commons</groupId>
+    <artifactId>sonar-analyzer-commons-parent</artifactId>
+    <version>1.5-SNAPSHOT</version>
```
pynicolas: We will have to increment that if we don't want to merge this PR before the release of 1.5.

m-g-sonar: Oh, good point!

PR #45, opened 2018-03-20
Issue#42 Replace use of deprecated RuleFinder by BuiltInQualityProfilesDefinition

christophe-zurn-sonarsource:


pom.xml
```diff
@@ -64,7 +64,7 @@
 
   <properties>
     <!-- versions -->
-    <version.sonar>6.4</version.sonar>
+    <version.sonar>6.7</version.sonar>
```
pynicolas: I think it would be better to create a separate ticket for this little change.

commons/src/main/java/org/sonarsource/analyzer/commons/BuiltInQualityProfileReader.java
```diff
@@ -0,0 +1,65 @@
...
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.sonar.api.server.profile.BuiltInQualityProfilesDefinition.NewBuiltInQualityProfile;
+
+/**
+ * Use to create {@link NewBuiltInQualityProfile} based on json file, for profiles implementing {@link org.sonar.api.server.profile.BuiltInQualityProfilesDefinition}
+ *
+ * Not designed for multi-threads
+ */
+public final class BuiltInQualityProfileReader {
```
pynicolas: I think it would be more explicit to name it something like `JsonBuiltInQualityProfile`.

commons/src/main/java/org/sonarsource/analyzer/commons/BuiltInQualityProfileReader.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private NewBuiltInQualityProfile profile;
+
+  public BuiltInQualityProfileReader(NewBuiltInQualityProfile profile) {
```
pynicolas: The separation between the constructor and `activateRulesFromJson` seems a bit artificial. Maybe we could move the `profile` parameter to `activateRulesFromJson`and make `activateRulesFromJson` static.

pom.xml
christophe-zurn-sonarsource: `BuiltInQualityProfilesDefinition` was introduced in SQ API 6.6, so the other changes are based on this update. But I can make a separate commit for it so that we keep a trace in commit history

commons/src/main/java/org/sonarsource/analyzer/commons/BuiltInQualityProfileReader.java
christophe-zurn-sonarsource: Ok

commons/src/main/java/org/sonarsource/analyzer/commons/BuiltInQualityProfileReader.java
christophe-zurn-sonarsource: ok

PR #59, opened 2018-07-17
Update version in pom.xml (should have been done before)

pynicolas:


vilchik-elena:
argh... always forget to do that, thanks!:)

PR #60, opened 2018-07-18
Add support for security hotspots

andrea-guarino-sonarsource:


commons/src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java
```diff
@@ -59,6 +63,11 @@ public RuleMetadataLoader(String resourceFolder, String defaultProfilePath) {
     this(resourceFolder, BuiltInQualityProfileJsonLoader.loadActiveKeysFromJsonProfile(defaultProfilePath));
   }
 
+  public RuleMetadataLoader(String resourceFolder, String defaultProfilePath, SonarRuntime sonarRuntime) {
+    this(resourceFolder, BuiltInQualityProfileJsonLoader.loadActiveKeysFromJsonProfile(defaultProfilePath));
```
vilchik-elena: you don't need to compute `BuiltInQualityProfileJsonLoader.loadActiveKeysFromJsonProfile(defaultProfilePath)` as you can call previous constructor (and avoid duplication this way)

commons/src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java
```diff
@@ -142,14 +151,31 @@ private void setMetadataFromJson(NewRule rule) {
...
     rule.setStatus(RuleStatus.valueOf(getUpperCaseString(root, "status")));
     rule.setTags(getStringArray(root, "tags"));
 
     Object remediation = root.get("remediation");
     if (remediation != null) {
       setRemediationFromJson(rule, (Map<String, Object>) remediation);
     }
+
+    Object securityStandards = root.get("securityStandards");
+    if (securityStandards != null && this.supportsSecurityHotspots) {
```
vilchik-elena: `this.` is redundant

commons/src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java
```diff
@@ -142,14 +151,31 @@ private void setMetadataFromJson(NewRule rule) {
...
+      setSecurityStandardsFromJson(rule, (Map<String, Object>) securityStandards);
+    }
+  }
+
+  private static void setSecurityStandardsFromJson(NewRule rule, Map<String, Object> securityStandards) {
+    for (String s : getStringArray(securityStandards, "OWASP")) {
```
vilchik-elena: can we rename `s` to `standard`?

commons/src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java
```diff
@@ -193,4 +219,17 @@ private static String getString(Map<String, Object> map, String propertyName) {
     return ((List<String>) propertyValue).toArray(new String[0]);
   }
 
+  static int[] getIntArray(Map<String, Object> map, String propertyName) {
```
vilchik-elena: should it be private?

commons/src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java
```diff
@@ -142,38 +149,43 @@ private void setDescriptionFromHtml(NewRule rule) {
...
+    if (securityStandards != null && supportsSecurityHotspots) {
       setSecurityStandardsFromJson(rule, (Map<String, Object>) securityStandards);
     }
   }
 
+  Map<String, Object> getMetadata(String ruleKey) {
+    String jsonPath = resourceFolder + RESOURCE_SEP + ruleKey + ".json";
+    Map<String, Object> metadata;
+    try {
+      metadata = jsonParser.parse(Resources.toString(jsonPath, UTF_8));
```
vilchik-elena: you might just return here and not create variable here

PR #75, opened 2018-12-20
xml-parsing: add secondary with message

vilchik-elena:


xml-parsing/src/main/java/org/sonarsource/analyzer/commons/xml/checks/SonarXmlCheck.java
```diff
@@ -90,4 +97,19 @@ public final void reportIssue(Node node, String message) {
     XmlTextRange textRange = XmlFile.nodeLocation(node);
     reportIssue(textRange, message, Collections.emptyList());
   }
+
+  public static class Secondary {
+    final XmlTextRange range;
+    final String message;
```
m-g-sonar: This should also be annotated `@Nullable`

vilchik-elena: Why? there is no use of that. The only useful thing, is documentation purpose.

PR #77, opened 2018-12-21
Add more secure parsing features

vilchik-elena:


xml-parsing/src/main/java/org/sonarsource/analyzer/commons/xml/SafetyFactory.java
```diff
@@ -39,6 +39,7 @@ public static XMLInputFactory createXMLInputFactory() {
     factory.setProperty(XMLInputFactory.SUPPORT_DTD, false);
     factory.setProperty(XMLInputFactory.IS_REPLACING_ENTITY_REFERENCES, false);
     factory.setProperty(XMLInputFactory.IS_VALIDATING, false);
+    factory.setProperty("javax.xml.stream.isSupportingExternalEntities", false);
```
m-g-sonar: I would use `XMLInputFactory.IS_SUPPORTING_EXTERNAL_ENTITIES` instead.

PR #81, opened 2019-01-03
Update XML dependencies to latest versions

m-g-sonar:


xml-parsing/pom.xml
```diff
@@ -40,15 +40,23 @@
...
     </dependency>
 
+    <!-- Provide safe DocumentBuilder to parse XML files -->
     <dependency>
       <groupId>xerces</groupId>
       <artifactId>xercesImpl</artifactId>
-      <version>2.8.1</version>
+      <version>2.12.0</version>
     </dependency>
+    <!-- required by xerces, as it provide some interfaces for xml nodes -->
```
vilchik-elena: it provide**s**

xml-parsing/pom.xml
```diff
@@ -40,15 +40,23 @@
       <version>1.11-SNAPSHOT</version>
     </dependency>
 
+    <!-- Provide safe DocumentBuilder to parse XML files -->
```
vilchik-elena: what is this comment for?

PR #82, opened 2019-01-03
Define simple XPath based abstract check

m-g-sonar:


xml-parsing/src/main/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheck.java
```diff
@@ -0,0 +1,93 @@
...
+import org.sonar.api.utils.log.Loggers;
+import org.w3c.dom.NamedNodeMap;
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+public abstract class SimpleXPathBasedCheck extends SonarXmlCheck {
+
+  private static final Logger LOG = Loggers.get(SimpleXPathBasedCheck.class);
+
+  private final XPath xpath = XPathFactory.newInstance().newXPath();
```
vilchik-elena: can we use constructor instead?

xml-parsing/src/main/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheck.java
```diff
@@ -0,0 +1,93 @@
...
+  public XPathExpression getXPathExpression(String expression) {
+    try {
+      return xpath.compile(expression);
+    } catch (XPathExpressionException e) {
+      throw new IllegalStateException(String.format("[%s] Fail to compile XPath expression '%s'.", ruleKey(), expression), e);
+    }
+  }
+
+  @CheckForNull
+  public NodeList evaluate(XPathExpression expression, Node node) {
```
vilchik-elena: how about adding a doc explaining when result is null?

xml-parsing/src/main/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheck.java
```diff
@@ -0,0 +1,93 @@
...
+import org.w3c.dom.Node;
+import org.w3c.dom.NodeList;
+
+public abstract class SimpleXPathBasedCheck extends SonarXmlCheck {
+
+  private static final Logger LOG = Loggers.get(SimpleXPathBasedCheck.class);
+
+  private final XPath xpath = XPathFactory.newInstance().newXPath();
+
+  public XPathExpression getXPathExpression(String expression) {
```
vilchik-elena: how about adding a doc explaining when exception is thrown by the method?

xml-parsing/src/main/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheck.java
```diff
@@ -0,0 +1,93 @@
...
+    if (numberResults == 0) {
+      return Collections.emptyList();
+    }
+    return IntStream.range(0, numberResults)
+      .mapToObj(nodeList::item)
+      .collect(Collectors.toList());
+  }
+
+  @CheckForNull
+  public static Node nodeAttribute(Node node, String attribute) {
```
vilchik-elena: I suggest to move this one to `XmlFile` class

xml-parsing/src/main/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheck.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @CheckForNull
+  public NodeList evaluate(XPathExpression expression, Node node) {
+    try {
+      return (NodeList) expression.evaluate(node, XPathConstants.NODESET);
+    } catch (XPathExpressionException e) {
+      if (LOG.isDebugEnabled()) {
+        RuleKey ruleKey = ruleKey();
+        LOG.debug(String.format("[%s] Unable to evaluate XPath expression on file %s", ruleKey, inputFile()));
+        LOG.error(String.format("[%s] XPath exception:", ruleKey), e);
```
vilchik-elena: how about simplifying this to:
```
if (LOG.isDebugEnabled()) {
    RuleKey ruleKey = ruleKey();
    LOG.error(String.format("[%s] Unable to evaluate XPath expression on file %s", ruleKey, inputFile()), e);
}
```

xml-parsing/src/test/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheckTest.java
```diff
@@ -0,0 +1,170 @@
...
+    SimpleXPathBasedCheck check = new SimpleXPathBasedCheck() {
+      XPathExpression bs = getXPathExpression("//b");
+      XPathExpression ds = getXPathExpression("//d");
+
+      @Override
+      protected void scanFile(XmlFile file) {
+        Document node = file.getDocument();
+
+        NodeList result = evaluate(bs, node);
+        assertThat(result).isNotNull();
```
vilchik-elena: is there a way to assert presence of assertions in the test? I think it's useful when assert is inside nested call like here.

xml-parsing/src/test/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheckTest.java
```diff
@@ -0,0 +1,170 @@
...
+        assertThat(empty.getLength()).isEqualTo(0);
+        List<Node> emptyAsList = evaluateAsList(ds, node);
+        assertThat(emptyAsList).isEmpty();
+
+        assertThat(asList(null)).isEmpty();
+      }
+    };
+
+    check.scanFile(null, RuleKey.of("tst", "nodeList"), xmlFile);
+    assertThat(logTester.logs()).isEmpty();
```
vilchik-elena: should we set logs to debug before?

xml-parsing/src/test/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheckTest.java
```diff
@@ -0,0 +1,170 @@
...
+        List<Node> asList = evaluateAsList(bs, node);
+        assertThat(asList).hasSize(4);
+
+        NodeList empty = evaluate(ds, node);
+        assertThat(empty).isNotNull();
+        assertThat(empty.getLength()).isEqualTo(0);
+        List<Node> emptyAsList = evaluateAsList(ds, node);
+        assertThat(emptyAsList).isEmpty();
+
+        assertThat(asList(null)).isEmpty();
```
vilchik-elena: I prefer to move this line out of this method, we don't need `scanFile` scope for it.

xml-parsing/src/test/resources/checks/SimpleXPathBasedCheck/xPathFailure.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!DOCTYPE foo [<!ELEMENT foo ANY >
+        <!ENTITY xxe SYSTEM "file:///dev/random" >]>
+
+<foo>&xxe;</foo>
```
vilchik-elena: it  will be nice to add comment here about how this file fails and that we don't know why.

xml-parsing/src/test/java/org/sonarsource/analyzer/commons/xml/checks/SimpleXPathBasedCheckTest.java
m-g-sonar: In fact I even moved `asList` to `XmlFile` class

PR #93, opened 2019-07-05
Prevent test related dependencies to be part of the runtime classpath

alban-auzeill:


vilchik-elena:
@alban-auzeill I think we should add size test to be sure about it in future

PR #100, opened 2019-09-04
Add annotation for addition of deprecated rule keys

vilchik-elena:


vilchik-elena:
Tibor, while adding docs I noticed that this deprecated rulekey api was added in SQ 7.1, so it will fail with older SQ. I decided to bump dependency for SQ to LTS, so that we don't need to deal with compatability.

xml-parsing/src/main/java/org/sonarsource/analyzer/commons/xml/checks/SonarXmlCheckVerifier.java
```diff
@@ -76,8 +75,7 @@ public static void verifyNoIssue(String relativePath, SonarXmlCheck check) {
   private static SonarXmlCheckVerifier createVerifier(String fileName, SonarXmlCheck check) {
     File file = new File(new File(BASE_DIR.toFile(), check.getClass().getSimpleName()), fileName);
 
-    SensorContextTester context = SensorContextTester.create(BASE_DIR)
-      .setActiveRules(new ActiveRulesBuilder().create(RULE_KEY).activate().build());
+    SensorContextTester context = SensorContextTester.create(BASE_DIR);
```
saberduck: why we don't need the rule to be activated anymore?

vilchik-elena: @saberduck good point, no idea. Probably SensorContextTester accepts issues even their rules are not activated. I think we didn't even check that when adding this line initially.

PR #103, opened 2019-11-11
Split SafetyFactory to avoid requiring xerces when using only stax

### LINKED ISSUE #107: Using the safe StAX parser shouldn't require embedding Xerces
Currently, the `SafetyFactory` class exposes 2 methods to parse XML: 1 for StAX and 1 for DOM. To parse DOM, we use Xerces.

Because of that, any class which depends on `SafetyFactory` also has some dependency on Xerces. SonarSource plugins which use `SafetyFactory` therefore embed Xerces even if don't need DOM and Xerces is about 1MB.

We can easily break this dependency by splitting `SafetyFactory` in 2 classes. That can save 1MB on the JAR file of each SonarSource plugin which uses `SafetyFactory` and doesn't use DOM.

pynicolas:
Fixes #107


m-g-sonar:
LGTM! Nice work!

m-g-sonar:
rebased today on top of master

PR #118, opened 2020-09-16
Hardening for SonarLint

henryju:


henryju:
@dbmeneses I would like your opinion on the first commit since it is an attempt to fix this very old issue: https://jira.sonarsource.com/browse/SSLRSQBR-72 and since it is about Thread, I would like your opinion.

dbmeneses:
LGTM!

commons/src/main/java/org/sonarsource/analyzer/commons/ProgressReport.java
```diff
@@ -54,22 +55,33 @@ public ProgressReport(String threadName, long period) {
...
-      if (success) {
-        log(count + "/" + count + " source files have been " + adjective);
-      }
+    if (success) {
+      log(count + "/" + count + " source " + pluralizeFile(count) + " " + pluralizeHas(count) + " been " + adjective);
+    }
+  }
+
+  private static String pluralizeFile(long count) {
+    if (count == 1) {
```
jblievremont, line 74: Nitpick (and same in `pluralizeHas`):
```suggestion
    if (count == 1L) {
```

jblievremont (approved):
:+1:

commons/src/main/java/org/sonarsource/analyzer/commons/ProgressReport.java
```diff
@@ -92,14 +103,20 @@ public synchronized void nextFile() {
...
+    join();
   }
 
-  public void join() throws InterruptedException {
-    thread.join();
+  private void join() {
+    try {
+      thread.join();
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
```
m-g-sonar, line 118: These two lines are apparently not covered. Would it be possible to try to check them?
I feel it's important as it propagates the interruption.

m-g-sonar:
@henryju If you can fix the coverage, I would appreciate it. I also think that small fix suggested by @jblievremont are good improvements.
Otherwise, LGTM!

commons/src/main/java/org/sonarsource/analyzer/commons/ProgressReport.java
henryju: Not very simple to test, but here it is.

m-g-sonar (approved):
LGTM! thanks for the changes!

PR #147, opened 2021-07-28
Import Regex helper classes from sonar-java

guillaume-dequenne-sonarsource:


regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/SimplifiedRegexCharacterClass.java
```diff
@@ -0,0 +1,367 @@
...
+
+  public void add(CharacterClassElementTree tree) {
+    new Builder(this).visitInCharClass(tree);
+  }
+
+  public void add(DotTree tree) {
+    char[] orderedExcludedCharacters;
+    if (tree.activeFlags().contains(Pattern.DOTALL)) {
+      orderedExcludedCharacters = new char[] {};
+    } else if (tree.activeFlags().contains(Pattern.UNIX_LINES)) {
```
sebastian-hungerecker-sonarsource, line 92: This is another point where behavior may differ between languages. According to some quick testing, Python at least will consider `\n` to be the only line break character regardless of which flags are set (and doesn't seem to have an equivalent for the `UNIX_LINES` flag).

sebastian-hungerecker-sonarsource:
I don't think the check for unknown characters when handling negated character classes in  `SimplifiedRegexCharacterClass ` should be removed. Properly tracking unknown characters is important to avoid false positives.

sebastian-hungerecker-sonarsource (approved):
LGTM

PR #149, opened 2021-08-06
Add file provider based on path pattern

yassin-kammoun-sonarsource:
Such file provided is useful user-provided reports (e.g. coverage)

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
@@ -38,7 +38,7 @@ public class PatternDirectoryScanner {
     this.pattern = pattern;
   }
 
-  public List<File> getIncludedFiles() {
+  public List<File> getMatchingFiles() {
     final String baseDirAbsolutePath = baseDir.getAbsolutePath();
     IOFileFilter fileFilter = new IOFileFilter() {
 
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
@@ -33,9 +33,9 @@ public class PatternDirectoryScanner {
   private final File baseDir;
   private final WildcardPattern pattern;
 
-  public PatternDirectoryScanner(File baseDir, WildcardPattern pattern) {
+  public PatternDirectoryScanner(File baseDir, String pattern) {
     this.baseDir = baseDir;
-    this.pattern = pattern;
+    this.pattern = WildcardPattern.create(pattern);
   }
 
   public List<File> getMatchingFiles() {
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
@@ -1,0 +1,60 @@
+/*
+ * SonarSource Analyzers Commons
+ * Copyright (C) 2009-2021 SonarSource SA
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.analyzer.commons;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.io.filefilter.IOFileFilter;
+import org.apache.commons.io.filefilter.TrueFileFilter;
+import org.sonar.api.utils.WildcardPattern;
+
+public class PatternDirectoryScanner {
+
+  private final File baseDir;
+  private final WildcardPattern pattern;
+
+  public PatternDirectoryScanner(File baseDir, WildcardPattern pattern) {
+    this.baseDir = baseDir;
+    this.pattern = pattern;
+  }
+
+  public List<File> getIncludedFiles() {
+    final String baseDirAbsolutePath = baseDir.getAbsolutePath();
+    IOFileFilter fileFilter = new IOFileFilter() {
+
+      @Override
+      public boolean accept(File dir, String name) {
+        return accept(new File(dir, name));
+      }
+
+      @Override
+      public boolean accept(File file) {
+        String path = file.getAbsolutePath();
+        path = path.substring(Math.min(baseDirAbsolutePath.length(), path.length()));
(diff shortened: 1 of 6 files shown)


commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
```diff
@@ -0,0 +1,60 @@
...
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.FilenameUtils;
+import org.apache.commons.io.filefilter.IOFileFilter;
+import org.apache.commons.io.filefilter.TrueFileFilter;
+import org.sonar.api.utils.WildcardPattern;
+
+public class PatternDirectoryScanner {
```
vilchik-elena, line 31: now I think if we should name it `DirectoryPatternProvider` :) seems to fit better

commons/src/test/java/org/sonarsource/analyzer/commons/PatternDirectoryScannerTest.java
```diff
@@ -0,0 +1,82 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class PatternDirectoryScannerTest {
+
+  private static final String F_1_TXT = "dir/f1.txt";
+
+  private final File baseDir = new File("src/test/resources/org/sonarsource/analyzer/commons/scanner").getAbsoluteFile();
```
vilchik-elena, line 38: why not static?

commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
```diff
@@ -0,0 +1,60 @@
...
+    this.pattern = WildcardPattern.create(pattern);
+  }
+
+  public List<File> getMatchingFiles() {
+    final String baseDirAbsolutePath = baseDir.getAbsolutePath();
+    IOFileFilter fileFilter = new IOFileFilter() {
+
+      @Override
+      public boolean accept(File dir, String name) {
+        return accept(new File(dir, name));
```
vilchik-elena, line 47: do you know how to cover this line?

commons/src/test/java/org/sonarsource/analyzer/commons/PatternDirectoryScannerTest.java
```diff
@@ -0,0 +1,82 @@
...
+      public File[] answer(InvocationOnMock invocation) throws Throwable {
+        FileFilter filter = (FileFilter) invocation.getArguments()[0];
+        filter.accept(new File("/short/path"));
+        return new File[] {matchingFile};
+      }
+    });
+    assertThat(scan("xxx", dir)).containsOnly(matchingFile);
+  }
+
+  private List<File> scan(String pattern) {
```
vilchik-elena, line 73: `getMatchingFiles`

commons/src/test/java/org/sonarsource/analyzer/commons/PatternDirectoryScannerTest.java
```diff
@@ -0,0 +1,82 @@
...
+  private final File baseDir = new File("src/test/resources/org/sonarsource/analyzer/commons/scanner").getAbsoluteFile();
+
+  @Test
+  public void noMatchedFile() {
+    assertThat(scan("dir/xxx")).isEmpty();
+  }
+
+  @Test
+  public void simpleFile() {
+    assertThat(scan(F_1_TXT)).containsOnly(new File(baseDir, F_1_TXT));
```
vilchik-elena, line 47: I would use literal instead of constant as `scan` argument

commons/src/test/java/org/sonarsource/analyzer/commons/PatternDirectoryScannerTest.java
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  @Test
+  public void wildCard() {
+    assertThat(scan("*/f1.txt")).containsOnly(new File(baseDir, F_1_TXT));
```
vilchik-elena, line 52: how about creating constants for those 2 files

commons/src/test/java/org/sonarsource/analyzer/commons/PatternDirectoryScannerTest.java
```diff
@@ -0,0 +1,82 @@
...
+import org.mockito.stubbing.Answer;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class PatternDirectoryScannerTest {
+
+  private static final String F_1_TXT = "dir/f1.txt";
```
vilchik-elena, line 36: this constant is confusing, why it's referencing first file, while second one has the same name as this variable name

commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
```diff
@@ -0,0 +1,60 @@
...
+      }
+
+      @Override
+      public boolean accept(File file) {
+        String path = file.getAbsolutePath();
+        path = path.substring(Math.min(baseDirAbsolutePath.length(), path.length()));
```
vilchik-elena, line 53: do you know why we need this check? is it just defensive programming? I don't see when it can happen

vilchik-elena (changes requested):

commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
yassin-kammoun-sonarsource: Yes, I can modify a test case for that.

commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
yassin-kammoun-sonarsource: You will have to convince me because it sounds weird if I try to translate that class name into a sentence: `this class is a provider of pattern directories`. To be honest, I actually liked the original name the Python bubble came up with.

vilchik-elena: But scanner should have `scan` method with some behaviour applied to each scanned item. But I agree that my option is also not perfect :)

commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
yassin-kammoun-sonarsource: The original [commit](https://github.com/SonarSource/sonar-python/commit/d650edab79cad414514b405a88f7f356d8fb185f#diff-0ad39d8ead8e6fb85c8667610e59d773717e8008ab608d970d11ae5dda8dfdc7) doesn't really help to understand why, but judging by the following test case, I have a feeling it's either due to unexpected behaviours on some operating systems or to handle symbolic links :

https://github.com/SonarSource/sonar-analyzer-commons/blob/dda3d0a7c077cd0cad0b2d96ac799c9acbfc9dde/commons/src/test/java/org/sonarsource/analyzer/commons/PatternDirectoryScannerTest.java#L56-L71

commons/src/main/java/org/sonarsource/analyzer/commons/PatternDirectoryScanner.java
yassin-kammoun-sonarsource: I think you are a bit influenced by the semantics (and API) of scanners at SonarSource :), and now that I think about it, it does make sense not to use the term "scanner" here to avoid the confusion.

What about `FileProvider` then ?

vilchik-elena: sounds good

vilchik-elena (approved):

PR #181, opened 2021-11-18
Bump woodstox-core and gson version

dorian-burihabwa-sonarsource:


saberduck:
@dorian-burihabwa-sonarsource can't we update to the latest available version 6.2.6? It seems 5.3.0 is still quite old

dorian-burihabwa-sonarsource:
> @dorian-burihabwa-sonarsource can't we update to the latest available version 6.2.6? It seems 5.3.0 is still quite old

Trying now 6.2.6.

saberduck (approved):

PR #188, opened 2021-11-30
RuleMetadataLoader: use new SQ API to provide OWASP 2021 standards

vilchik-elena:


commons/src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java
```diff
@@ -177,10 +183,18 @@ private void setMetadataFromJson(NewRule rule) {
     }
   }
 
-  private static void setSecurityStandardsFromJson(NewRule rule, Map<String, Object> securityStandards) {
-    if (securityStandards.get("OWASP") != null) {
-      for (String standard : getStringArray(securityStandards, "OWASP")) {
-        rule.addOwaspTop10(RulesDefinition.OwaspTop10.valueOf(standard));
+  private void setSecurityStandardsFromJson(NewRule rule, Map<String, Object> securityStandards) {
+    boolean isOwaspByVersionSupported = sonarRuntime.getApiVersion().isGreaterThanOrEqual(Version.create(9, 3));
+    if (isOwaspByVersionSupported) {
```
vilchik-elena, line 188: @nils-werner-sonarsource I think we should have `else` branch here falling into deprecated `addOwaspTop10` call when we are on SQ <9.3
I believe it might be theoretically possible (while it's hard to think of such case, may be in SonarLint context) when we want to update sonar-analyzer-commons on a plugin relying on older version of SQ.

nils-werner-sonarsource (approved):
LGTM!

PR #196, opened 2022-01-26
Resolve #195 Implement VerboseRegexFinder for Rule S6353

### LINKED ISSUE #195: Implement VerboseRegexFinder for Rule S6353
Related to [SONARPHP-1276](https://jira.sonarsource.com/browse/SONARPHP-1276)

nils-werner-sonarsource:


regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/VerboseRegexFinder.java
```diff
@@ -0,0 +1,167 @@
...
+    CharacterClassElementTree element = tree.getContents();
+    if (element.is(CHARACTER_RANGE) && "0-9".equals(element.getText())) {
+      String expected = backslash(tree) + (tree.isNegated() ? "D" : "d");
+      reportVerboseCharacterClass(expected, tree);
+    }
+  }
+
+  private void checkBulkyAnyCharacterClass(CharacterClassTree tree) {
+    CharacterClassElementTree element = tree.getContents();
+    if (tree.isNegated() || !element.is(UNION) || ((CharacterClassUnionTree) element).getCharacterClasses().size() != 2) {
```
stanislav-jordanov-sonarsource, line 100: I was gonna ask "What about `/[\w\Wx]/` ?"
But then this is probably going to fire another rule, stating that the 'x' is redundant, and when 'x' is removed the rest will lend in this case?

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/VerboseRegexFinder.java
```diff
@@ -0,0 +1,167 @@
...
+
+  private void reportVerboseCharacterClass(String expected, CharacterClassTree tree) {
+    String message = String.format(VERBOSE_CHARACTER_CLASS_MESSAGE, expected, tree.getText());
+    regexElementIssueReporter.report(tree, message, null, Collections.emptyList());
+  }
+
+  private void checkBulkyQuantifier(Quantifier quantifier) {
+    String raw = quantifier.getText();
+    BulkyQuantifier bulkyQuantifier = null;
+    if (Pattern.matches("\\{0,1}\\??$", raw)) {
```
stanislav-jordanov-sonarsource, line 139: I am sure this is correct, I just don't understand:
1) why the double slash around the curly braces?
2) what does the double question mark (??) in the end means?

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/VerboseRegexFinder.java
```diff
@@ -0,0 +1,167 @@
...
+      bulkyQuantifier = new BulkyQuantifier("?", "{0,1}");
+    } else if (Pattern.matches("\\{0,}\\??", raw)) {
+      bulkyQuantifier = new BulkyQuantifier("*", "{0,}");
+    } else if (Pattern.matches("\\{1,}\\??$", raw)) {
```
stanislav-jordanov-sonarsource, line 143: I don't get the optional question mark at the end - even more so because /X+?/ is not equivalent to /X+/ but to /X*/
Is it there to designate a reluctant match?

regex-parsing/src/test/resources/finders/VerboseRegexFinder.yml
```diff
@@ -0,0 +1,39 @@
...
+- '[A-Za-z0-9_]'      # Noncompliant {{Use concise character class syntax '\w' instead of '[A-Za-z0-9_]'.}}
+- '[0-9_A-Za-z]'      # Noncompliant
+- '[^A-Za-z0-9_]'     # Noncompliant {{Use concise character class syntax '\W' instead of '[^A-Za-z0-9_]'.}}
+- '[^0-9_A-Za-z]'     # Noncompliant
+- 'x{0,1}'            # Noncompliant {{Use concise quantifier syntax '?' instead of '{0,1}'.}}
+- 'x{0,1}?'           # Noncompliant
+- 'x{0,}'             # Noncompliant {{Use concise quantifier syntax '*' instead of '{0,}'.}}
+- 'x{0,}?'            # Noncompliant
+- 'x{1,}'             # Noncompliant {{Use concise quantifier syntax '+' instead of '{1,}'.}}
+- 'x{1,}?'            # Noncompliant
```
stanislav-jordanov-sonarsource, line 17: There!
The message expected is not quoted, but is it not true that /x{0,1}?/ is not the same as /x+/ ?
Or does the `?` in this case mean something else instead of optional?

stanislav-jordanov-sonarsource (approved):
With few questions and one possible issue: LGTM.

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/VerboseRegexFinder.java
nils-werner-sonarsource: It will be covered by the rule which detects duplications in character classes

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/VerboseRegexFinder.java
nils-werner-sonarsource: It's a lazy quantifier with the question mark
See https://www.regular-expressions.info/repeat.html#lazy

regex-parsing/src/test/resources/finders/VerboseRegexFinder.yml
nils-werner-sonarsource: See lazy quantifiers

PR #204, opened 2022-02-01
Implement EmptyGroupFinder for Rule 6331

stanislav-jordanov-sonarsource:


regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/ast/GroupTree.java
```diff
@@ -30,7 +30,8 @@
   private final RegexTree.Kind kind;
 
   /**
-   * Can only be null for non-capturing groups
+   * Can only be null for non-capturing groups (by design).
+   * If non-null - it represents an empty group iff `element` is a SequenceTree with empty `items` list.
```
nils-werner-sonarsource, line 34:
```suggestion
   * If non-null - it represents an empty group if `element` is a SequenceTree with empty `items` list.
```

nils-werner-sonarsource (approved):
LGTM! See the typo and please add the missing test case to get 100% coverage.

PR #206, opened 2022-02-02
fix EmptyGroupFinder so it does not report modifiers as an empty group

stanislav-jordanov-sonarsource:
Modifiers are represented in the AST as NonCapturingGroup with element = null.
Don't treat those like empty-non capturing groups (they are not!) and don't raise S6331


nils-werner-sonarsource (approved):
LGTM!

PR #207, opened 2022-02-02
Report on orOperator in EmptyAlternativeFinder

nils-werner-sonarsource:


regex-parsing/src/test/java/org/sonarsource/analyzer/commons/regex/finders/RegexFinderVerifier.java
```diff
@@ -168,16 +168,24 @@ public void reportInvocationTreeIssue(Node regexNode, String message, @Nullable
...
       secondaries.forEach(location -> {
         IssueLocation.Range secondaryRange = issueLocationInFile(location);
+        verifyHighlightableRange(range);
         issue.addSecondary(secondaryRange.getLine(), secondaryRange.getColumn(), secondaryRange.getEndLine(),
           secondaryRange.getEndColumn(), location.message());
       });
     }
 
+    private void verifyHighlightableRange(IssueLocation.Range range) {
+      if (range.getLine() == range.getEndLine() && range.getColumn() == range.getEndColumn()) {
```
stanislav-jordanov-sonarsource, line 184: I am not sure I understand what exactly this condition means.

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/EmptyAlternativeFinder.java
```diff
@@ -41,10 +42,17 @@ public EmptyAlternativeFinder(RegexIssueReporter.ElementIssue regexElementIssueR
   @Override
   public void visitDisjunction(DisjunctionTree tree) {
     List<RegexTree> alternatives = tree.getAlternatives();
-    alternatives.stream().limit((alternatives.size() - 1))
-        .filter(alternative -> alternative.is(RegexTree.Kind.SEQUENCE))
-        .filter(alternative -> ((SequenceTree) alternative).getItems().isEmpty())
-        .forEach(alternative -> regexElementIssueReporter.report(alternative, MESSAGE, null, Collections.emptyList()));
+    for (int i = 0; i < alternatives.size() - 1; i++) {
+      if (isEmptyAlternative(alternatives.get(i))) {
+        SourceCharacter orOperator = tree.getOrOperators().get(i);
```
stanislav-jordanov-sonarsource, line 47: If I understand correctly, here you highlight the orOperator (the '|' character) to the right of the empty alternative?

stanislav-jordanov-sonarsource (approved):

regex-parsing/src/test/java/org/sonarsource/analyzer/commons/regex/finders/RegexFinderVerifier.java
nils-werner-sonarsource: The parser creates trees, especially for an empty element which have a range of 0. In other words, the range of a tree could be from column 11 to column 11. This is a range inside the code that can not be highlighted aka displayed in the UI.

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/EmptyAlternativeFinder.java
nils-werner-sonarsource: We should be careful to report issues on trees that have no range. This verification inside the check verifier will show us at an early stage when we try to highlight a tree without range.

PR #209, opened 2022-02-04
Introduce a 'generic' visitGroupTree(GroupTree tree) method

stanislav-jordanov-sonarsource:
SONARPHP-1283 made us agree on the need to get rid of 4 * visitXyzGroup() methods (see EmptyAlternativeFined).
This PR solves this issue.


stanislav-jordanov-sonarsource:
Wait! There's a better way to do that ...

nils-werner-sonarsource (approved):
Thanks for your change. I spend some time formulating that the try-catch solution is quite ugly. Now it's clean and short. So far I remember we have another finder who could benefit from this new method. Maybe update this finder as well. All in all LGTM.

PR #211, opened 2022-02-07
Update parent pom and license header

nils-werner-sonarsource:


stanislav-jordanov-sonarsource (commented):
292 occurrences of "2009-2021"

PR #251, opened 2022-08-17
Allow iteration over PMap entries (#250)

### LINKED ISSUE #250: PMap should be iterable
At the time of writing, `PMap`s can be iterated only via the forEach method. `PSet` and `PStack` already implement `Iterable`.
Goal of this ticket is to make `PMap` implement the `Iterable` interface as well, such that it can be iterated via a regular `for`.

johann-beleites-sonarsource:
Fixes #250

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/collections/AVLTree.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/collections/AVLTree.java
@@ -19,11 +19,9 @@
  */
 package org.sonarsource.analyzer.commons.collections;
 
-import java.util.ArrayDeque;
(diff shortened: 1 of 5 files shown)


commons/src/main/java/org/sonarsource/analyzer/commons/collections/AVLTree.java
```diff
@@ -34,7 +32,7 @@
 
 /**
  * AVL Tree.
- *
+ * <p>
```
andreaguarino, line 35: Was this change intended?

andreaguarino (approved):
LGTM! Nice work!
Just a small comment about a change in a comment

PR #252, opened 2022-08-18
Read scope from rule json metadata file.

francoismora:
Fixes #252

### DIFF
--- a/commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
+++ b/commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
@@ -22,6 +22,7 @@ package org.sonarsource.analyzer.commons;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
```diff
@@ -233,6 +235,34 @@ public void getStringArray() throws Exception {
...
+  @Test
+  public void getScopeIfPresent() {
+    RuleScope scope;
+
+    assertThat(RuleMetadataLoader.getScopeIfPresent(Collections.emptyMap(), "scope")).isNotPresent();
+    assertThat(RuleMetadataLoader.getScopeIfPresent(Collections.singletonMap("scope", null), "scope")).isNotPresent();
+    assertThat(RuleMetadataLoader.getScopeIfPresent(Collections.singletonMap("scope", false), "scope")).isNotPresent();
+
+    scope = RuleMetadataLoader.getScopeIfPresent(Collections.singletonMap("scope", "main"), "scope").orElse(null);
+    assertThat(scope).isEqualTo(RuleScope.MAIN);
```
saberduck, line 247: I think `orElse(null)` is unnecessary, you can use `assertThat(scope).contains(...)` to assert directly optional value

commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
```diff
@@ -233,6 +235,34 @@ public void getStringArray() throws Exception {
     assertThat(RuleMetadataLoader.getStringArray(map, "key")).containsExactly("x", "y");
   }
 
+  @Test
+  public void getScopeIfPresent() {
+    RuleScope scope;
+
+    assertThat(RuleMetadataLoader.getScopeIfPresent(Collections.emptyMap(), "scope")).isNotPresent();
+    assertThat(RuleMetadataLoader.getScopeIfPresent(Collections.singletonMap("scope", null), "scope")).isNotPresent();
```
saberduck, line 243: Use static import for `singletonMap` and `emptyMap` to make code a bit more succint

saberduck (changes requested):

commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
francoismora: Thanks for pointing this out. I have updated accordingly.

commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
francoismora: Indeed I saw that too. The PR doesn't show that the same methods are used many times in the same file. There's undoubtedly an opportunity to refactor the class, for instance, to use the shorter form `List.of()`. I may do this as part of a separate PR, maybe? WDYT?

saberduck: we would need to change the `pom.xml` to use Java 11, otherwise `List.of` is not available. We could do this, but perhaps later.

saberduck (approved):

PR #257, opened 2022-08-30
Fix intersection/superset of lookahead and possessive quantifiers inside of a repetition

### LINKED ISSUE #256: Regex intesection/superset fails for lookaheads/possessive quantifiers inside a repetition
The current implementation fails to correctly check the intersection/superset of two `SubAutomaton` when used to test a lookahead/possessive quantifier inside a repetition. This impacts rules [S5994](https://rules.sonarsource.com/java/tag/regex/RSPEC-5994) and [S6002](https://rules.sonarsource.com/java/tag/regex/RSPEC-6002) as the following non-compliant examples are not detected:

- `(bx++)+x`
- `((?!ab)a)+b`

The current algorithm misses them because it takes into account the beginning of the sequence containing the negative lookahead/possessive quantifier. If the beginning of said sequence is not a subset of the lookahead/quantifier inner expression, false is returned. This not valid because the lookahead/quantifier will necessarily be encountered again, and so we are stuck in a loop.

To solve this issue, the algorithm has to ignore paths that always lead back to the lookahead/quantifier being tested.

Swalkyn:
Fix #256 

This change adds a way to exclude nodes of an Automaton from being transitioned to, based on their `IndexRange`. `BranchTrackingVisitor` keeps track of the last repetition/disjunction visited, which lets us exclude the correct nodes.

~~Performance-wise, it decreases the time needed for the corresponding rules (S5994 and S6002). However, rule S5855 also uses intersection but does not need the added feature, and takes a small performance hit.~~
Fixed by https://github.com/SonarSource/sonar-analyzer-commons/pull/257/commits/d895d2b2e449009b30de61598e8ccd89d8850a5a

### DIFF
--- a/regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/BranchTrackingVisitor.java
+++ b/regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/BranchTrackingVisitor.java
@@ -34,7 +34,7 @@ import org.sonarsource.analyzer.commons.regex.ast.RepetitionTree;
  * cycles during Automaton evaluation.
  */
 public class BranchTrackingVisitor extends RegexBaseVisitor {
-  private Deque<RegexTree> branchingNodes = new ArrayDeque<>();
+  private final Deque<RegexTree> branchingNodes = new ArrayDeque<>();
 
   @Override
   public void visitDisjunction(DisjunctionTree tree) {
--- a/regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/BranchTrackingVisitor.java
+++ b/regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/BranchTrackingVisitor.java
@@ -19,8 +19,6 @@
  */
 package org.sonarsource.analyzer.commons.regex.helpers;
 
-import java.util.ArrayDeque;
-import java.util.Deque;
 import org.sonarsource.analyzer.commons.regex.ast.AbstractRegexSyntaxElement;
 import org.sonarsource.analyzer.commons.regex.ast.DisjunctionTree;
 import org.sonarsource.analyzer.commons.regex.ast.IndexRange;
@@ -29,24 +27,26 @@ import org.sonarsource.analyzer.commons.regex.ast.RegexTree;
 import org.sonarsource.analyzer.commons.regex.ast.RepetitionTree;
 
 /**
- * The BranchTrackingVisitor saves the latest branching nodes as it traverses the tree. This is useful to avoid
- * cycles during Automaton evaluation.
+ * The BranchTrackingVisitor saves the nearest enclosing branching construct as it traverses the tree. This is useful to
+ * avoid cycles during Automaton evaluation.
  */
 public class BranchTrackingVisitor extends RegexBaseVisitor {
-  private final Deque<RegexTree> branchingNodes = new ArrayDeque<>();
+  private RegexTree branchingNode = null;
 
   @Override
   public void visitDisjunction(DisjunctionTree tree) {
-    branchingNodes.push(tree);
+    RegexTree previousBranchingNode = branchingNode;
+    branchingNode = tree;
     super.visitDisjunction(tree);
-    branchingNodes.pop();
+    branchingNode = previousBranchingNode;
   }
 
   @Override
   public void visitRepetition(RepetitionTree tree) {
-    branchingNodes.push(tree);
+    RegexTree previousBranchingNode = branchingNode;
+    branchingNode = tree;
     super.visitRepetition(tree);
-    branchingNodes.pop();
+    branchingNode = previousBranchingNode;
   }
 
   /**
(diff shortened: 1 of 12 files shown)


sebastian-hungerecker-sonarsource (changes requested):
After thinking about it some more, I don't think this approach is a general solution to the problem. Specifically the assumption, that the part of the automaton that you want to ignore is always a straight sequence to the quantifier/lookahead in question, is flawed. It will break when the loop containing the quantifier/lookahead contains a branch before the quantifier/lookahead. As a simple example, consider the following (failing) test case:

```yaml
- '(?:(x|y)(?!bc))+bc' # Noncompliant
```

A general solution will probably involve quite a bit more work (I'm assuming it'd involve trying to find a feasible path to the end of the regex instead of stopping early when you find input fulfilling the lookahead).

I'll leave it up to you whether you want to implement the more general solution or not.

If not, that leaves two options for this PR: merge it as a band-aid for the cases where it does work and leave the ticket open with a comment outlining which cases have been fixed and which are still problematic; or discard the PR and update the ticket with a description of this approach and why it didn't work.

To me, this choice depends on whether we feel that this partial solution already provides value to users. I'd be wary about adding the additional mental concept of ignored ranges in the code that would likely not be part of the general solution unless we have a justifying example where we say that it provides value to the user that we're able to detect this case sooner rather than later.

We can discuss this more if you want.

Swalkyn:
Very well thought! Indeed, when a dis-junction is part of the lookaround/possessive quantifier branch, the current code fails.

However, I think this is simply an incorrect implementation and the underlying observation still works. I think it is good to have it written somewhere:

_When running the automaton to check whether a lookaround/possessive quantifier is not contradictory, any state from which the examined lookaround/quantifier will necessarily be reached has to be excluded._

The problem is that, until now, I excluded the branch from the last visited dis junction/repetition. This is incorrect. What we really need is the branch from the last **ancestor** dis junction/repetition. Therefore, fixing `BranchTrackingVistor` to get this behavior will work.

I think this problem stemmed from a confusion I had with the visitor pattern. The way context is kept as the tree is traversed differs from when you use recursion and pattern matching on nodes (which is what I am used to). Example

```
class Visitor {
  closestAncestorBranch = null

  def visitBranch(Node branch) {
    closestBranch = branch
    branch.children.forEach(visit)
  }

  def visitLeaf(Node leaf) {
    ...
  }
```
does not behave like
```
def traverse(Node node, Branch closestAncestorBranch) {
  match (node) {
    case Branch(children) {
       children.forEach(child -> traverse(child, node)
    }
    case Leaf {}
  }
}
```
since `closestBranch` is not restored as we return. That's an important thing I am now aware of :)

If you can look at my fix when you have time, that would be great thanks! And if you still not convinced, I will be glad to try proving the idea formally.

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/BranchTrackingVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * The BranchTrackingVisitor saves the latest branching nodes as it traverses the tree. This is useful to avoid
+ * cycles during Automaton evaluation.
+ */
+public class BranchTrackingVisitor extends RegexBaseVisitor {
+  private final Deque<RegexTree> branchingNodes = new ArrayDeque<>();
+
+  @Override
+  public void visitDisjunction(DisjunctionTree tree) {
+    branchingNodes.push(tree);
```
sebastian-hungerecker-sonarsource, line 40: You don't actually need an explicit stack to restore state in visitors. Visitors still work recursively, so you still have the call stack to save and restore state. You just don't have the argument passing, so you need to use local variables to remember the field's old value and then restore it after.

So you can change `branchingNode` back to type `RegexTree`, replace the push with `RegexTree previousBranchingNode = branchingNode` and then the pop with `branchingNode = previousBranchingNode`.

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/BranchTrackingVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Deque;
+import org.sonarsource.analyzer.commons.regex.ast.AbstractRegexSyntaxElement;
+import org.sonarsource.analyzer.commons.regex.ast.DisjunctionTree;
+import org.sonarsource.analyzer.commons.regex.ast.IndexRange;
+import org.sonarsource.analyzer.commons.regex.ast.RegexBaseVisitor;
+import org.sonarsource.analyzer.commons.regex.ast.RegexTree;
+import org.sonarsource.analyzer.commons.regex.ast.RepetitionTree;
+
+/**
+ * The BranchTrackingVisitor saves the latest branching nodes as it traverses the tree. This is useful to avoid
```
sebastian-hungerecker-sonarsource, line 32: I think this description could be more clear. Until your fix it wasn't clear to me that the old behavior wasn't what you had in mind when you said "latest branching node". I think something like "nearest *enclosing* branching construct" would capture what you mean more closely.

sebastian-hungerecker-sonarsource (changes requested):
I think the fixed approach is general enough to be worthwhile, but there's at least one case that I don't think it can support: When two lookarounds' (or possessive quantifiers') infeasibility depend on each other. Consider this example:

    (xa(?=x)|xb(?=x))*c

Here the first lookahead can be fulfilled by running into the second lookahead and then the second can be fulfilled by running into the first, but you'll never be able to exit the loop this way.

I think that's an acceptable limitation for now though unless you feel like taking a crack at it. If not, we can merge this PR after addressing the two small comments I've left.

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/BranchTrackingVisitor.java
Swalkyn: Yep, of course

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/helpers/BranchTrackingVisitor.java
Swalkyn: Agreed, I changed it.

Swalkyn:
Yes, this does not work when you allow multiple lookarounds in the expressions. However, I believe supporting multiple lookarounds will require bigger changes. This was out of the scope of this work.

As you said, I think we can merge.

sebastian-hungerecker-sonarsource (approved):
LGTM. Can you create a ticket for the remaining limitation about multiple lookaheads and/or possessive quantifiers?

Swalkyn:
Issue describing the limitation: #258

PR #259, opened 2022-10-13
Remove dependency on commons-lang and commons-io

saberduck:
Remove dependency on commons-lang and commons-io, so they don't propagate transitively into other projects.

Please keep the commits separate (rebase)


alban-auzeill (approved):

saberduck:
Fixes #261

PR #268, opened 2022-10-28
Revert "Fix #256 Regex intesection/superset fails for lookaheads/possessive quantifiers inside a repetition (#257)"

### LINKED ISSUE #256: Regex intesection/superset fails for lookaheads/possessive quantifiers inside a repetition

alban-auzeill:
This reverts commit 645dd134cd3693d6455a4f789af28478d7170d22.


alban-auzeill:
We decided to not revert and fix the problem.

PR #269, opened 2022-10-28
Fix FPs in S5994 and S6002

Swalkyn:
PR #257 added support for lookahead and possessive quantifiers inside of repetitions by excluding nodes on the same branch from being traversed. This correctly prevented from exploring nodes that would lead back to the lookahead/possessive quantifier under analysis. However, this original solution was mistakenly excluding valid nodes, that came after the lookahead/quantifier.

This fix reduces the set of excluded node from a whole branch to only the part of the branch up to the lookahead/pos. quantifier itself.


dorian-burihabwa-sonarsource (approved):
LGTM!

PR #272, opened 2022-11-03
Add support for O(1) size query on AVLTree

gyula-sallai-sonarsource:


andreaguarino (approved):
LGTM. I think PStack and its implementation SingleLinkedList has `size()` method with O(n) comlpexity. Could you please create a GH issue to make it O(1) as well?

Godin (changes requested):
Sorry, but this implementation does not pass following test:

```
    Object k1 = new Key(42, "k1");
    Object k2 = new Key(42, "k2");
    AVLTree<Object, Object> t = AVLTree.create()
      .put(k1, "v1")
      .put(k2, "v2");
    assertThat(t.size())
      .isEqualTo(2);
```

sallaigy:
I am closing this for now: the implementation is not complete (see the comment above) and we no longer need it for DBD. We can always reopen in case we need some functionality like this in the future.

PR #277, opened 2023-02-21
MMF-3061 Provide extended rule description for some JS rules

### LINKED ISSUE #276: RuleMetadataLoader supports the education format for rules description
The `RuleMetadataLoader` class should be updated so that it can load a rule description written with either the standard(old) format, or the new education(new) format.
Most of the logic should be taken from the current implementation present in sonar-security, and adapted so that the `RuleMetadataLoader` will be backward compatible with the old `RulesDefinition.NewRule#setHtmlDescription` API.

christophe-zurn-sonarsource:
Fixes #276


commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
```diff
@@ -89,6 +89,51 @@ public void load_rule_S100() throws Exception {
...
+    class TestRule {
+    }
+    ruleMetadataLoader = new RuleMetadataLoader(RESOURCE_FOLDER, SONAR_RUNTIME_9_9);
+    ruleMetadataLoader.addRulesByAnnotatedClass(newRepository, singletonList(TestRule.class));
+    newRepository.done();
+
+    RulesDefinition.Repository repository = context.repository(RULE_REPOSITORY_KEY);
+    RulesDefinition.Rule rule = repository.rule("S102");
+    assertThat(rule).isNotNull();
+    assertThat(rule.htmlDescription().replaceAll("\r\n", "\n")).isEqualTo("Intro\n" +
```
ilia-kebets-sonarsource, line 126: The expected string is the same as in the test above. We could either modularize the expected string or make another assertion.

commons/src/test/java/org/sonarsource/analyzer/commons/EducationRuleLoaderTest.java
```diff
@@ -0,0 +1,275 @@
...
+    newRepository.done();
+    RulesDefinition.Rule rule = context.repository(RULE_REPOSITORY_KEY).rule("MyRuleKey");
+
+    assertThat(rule.ruleDescriptionSections()).isEmpty();
+    assertThat(fallbackDescription).isEqualTo(testFileContent);
+  }
+
+  @Test
+  public void education_description_content_unsupported_product_runtime_invalid_format() throws IOException {
+    SonarRuntime invalidRuntime = SonarRuntimeImpl.forSonarQube(Version.create(9, 4), SonarQubeSide.SERVER, SonarEdition.DEVELOPER);
```
ilia-kebets-sonarsource, line 103: maybe use a supported version for this case, to make sure the behaviour is triggered by the invalidity?

commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
christophe-zurn-sonarsource: Good point. I added the expected fallback string in a separate file, while the original has actually 2 frameworks, and the fallback only expects a single one :)

commons/src/test/java/org/sonarsource/analyzer/commons/EducationRuleLoaderTest.java
christophe-zurn-sonarsource: There's actually already another test that check that file with a supported version of the runtime.
This test was actually not useful, as it didn't test anything specific. I removed it :) :+1:

PR #278, opened 2023-02-23
Incorporate education rule loading into RuleMetadataLoader and introduce fallback mechanism

christophe-zurn-sonarsource:


christophe-zurn-sonarsource:
As discussed with @ilia-kebets-sonarsource , to ease review, I'm merging this PR into the feature branch

PR #286, opened 2023-03-21
SONARPY-1312 Introduce RedosFinder to mutualize implementation of S5852

guillaume-dequenne-sonarsource:


regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/RedosFinder.java
```diff
@@ -0,0 +1,324 @@
...
+        if (repetition.getQuantifier().getMinimumRepetitions() <= 1) {
+          collectSingleCharacters(repetition.getElement(), accumulator);
+        }
+      }
+      return accumulator;
+    }
+
+  }
+
+  private void addBacktracking(BacktrackingType newBacktrackingType) {
```
maksim-grebeniuk-sonarsource, line 260: it is easy to forget that ordering of enum elements is affects the logic - I would suggest to create enums field showing the ordering priority like:
```java
enum BacktrackingType implements Comparable<BacktrackingType> {
    ALWAYS_EXPONENTIAL(0),
    QUADRATIC_WHEN_OPTIMIZED(1),
    ALWAYS_QUADRATIC(2),
    LINEAR_WHEN_OPTIMIZED(3),
    NO_ISSUE(4);

    private final int priority;

    BacktrackingType(int priority) {
      this.priority = priority;
    }

    public BacktrackingType max(BacktrackingType another) {
      return this.priority > another.priority ? this : another;
    }

  }

```

and then use it in `addBacktracking` method like:
```java
private void addBacktracking(BacktrackingType newBacktrackingType) {
  foundBacktrackingType = foundBacktrackingType.max(newBacktrackingType);
}
```

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/RedosFinder.java
```diff
@@ -0,0 +1,324 @@
...
+    @Override
+    public void visitAtomicGroup(AtomicGroupTree tree) {
+      new RedosVisitor(tree, tree.continuation(), false, false).visit(tree);
+    }
+
+    @Override
+    public void visitRepetition(RepetitionTree tree) {
+      if (tree.isPossessive()) {
+        new RedosVisitor(tree, tree.continuation(), false, false).visit(tree);
+      } else if (containsIntersections(Arrays.asList(tree.getElement(), tree.continuation()))) {
```
maksim-grebeniuk-sonarsource, line 285: I'd suggest to use `List.of(...)`

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/RedosFinder.java
```diff
@@ -0,0 +1,324 @@
...
+        BacktrackingType greedyComplexity = tree.getQuantifier().isOpenEnded() ? BacktrackingType.QUADRATIC_WHEN_OPTIMIZED : BacktrackingType.LINEAR_WHEN_OPTIMIZED;
+        addBacktracking(isReluctant ? BacktrackingType.ALWAYS_EXPONENTIAL : greedyComplexity);
```
maksim-grebeniuk-sonarsource, line 287: I think there is if-else if block will be easier to understand the logic:
```java
var greedyComplexity = BacktrackingType.LINEAR_WHEN_OPTIMIZED;
if (isReluctant) {
  greedyComplexity = BacktrackingType.ALWAYS_EXPONENTIAL;
} else if (tree.getQuantifier().isOpenEnded()) {
  greedyComplexity = BacktrackingType.QUADRATIC_WHEN_OPTIMIZED
}
```

maksim-grebeniuk-sonarsource (approved):
Looks good, just it is better to add priority to enum

regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/finders/RedosFinder.java
guillaume-dequenne-sonarsource: To be honest I find the original snippet actually easier to understand, so I'll leave it like that if it's okay with you.

PR #297, opened 2023-06-12
Avoid calling 'scan' two times in code recognizers

Wohops:
Should improve the performance of the following detectors:
* `CamelCaseDetector`
* `ContainsDetector`
* `EndWithDetector`
* `KeywordsDetector`
* `RegexDetector`


sns-seb:
LGTM

I'm a bit concerned the test coverage is pretty light and wouldn't catch an unforeseen regression but the risks are super low from a code logic perspective

sns-seb (approved):

PR #298, opened 2023-06-23
Add `keySet` method on `PMap` and `stream` method on `PSet` and `PStack`

Swalkyn:
- Add a method to return the set of keys in a `PMap`
- Add a method to return a stream of the elements in a `PSet`
- Add a method to return a stream of the elements in a `PStack`, starting at the top of the stack


johann-beleites-sonarsource (approved):
LGTM!

PR #305, opened 2023-08-11
Prevent ExternalRuleLoader to manipulate code attribute and impact fields when runtime API < 10.1

alban-auzeill:


irina-batinic-sonarsource (approved):
LGTM!

PR #311, opened 2023-12-04
SONAR-21157 Fix CI error due to SQ 10.4 new requirement: ERROR EnvironmentInformation compiled 61 only recognizes up to 55

alban-auzeill:


dorian-burihabwa-sonarsource (approved):
Thank you for the change!

PR #321, opened 2024-04-23
SONARJAVA-4947 Add SingleFileVerifier.assertNoIssuesRaised

leonardo-pilastri-sonarsource:


test-commons/src/main/java/org/sonarsource/analyzer/commons/checks/verifier/SingleFileVerifier.java
```diff
@@ -99,6 +99,11 @@ static SingleFileVerifier create(Path sourceFilePath, Charset encoding) {
    */
   void assertNoIssues();
 
+  /**
+   * Run the comparison and expect to find no issue. Allows Noncompliant comments to be present in the file.
```
johann-beleites-sonarsource, line 103: Let's also update the comment of `assertNoIssues()` to explicitly mention that noncompliant comments are not allowed.

test-commons/src/main/java/org/sonarsource/analyzer/commons/checks/verifier/internal/InternalIssueVerifier.java
```diff
@@ -135,6 +135,12 @@ public void assertOneOrMoreIssues() {
     assertIssues(true);
   }
 
+  @Override
+  public void assertNoIssuesRaised() {
+    Report report = buildReport();
+    assertEquals("ERROR: Found " + report.getActualIssueCount() + " unexpected issues.", 0, report.getActualIssueCount());
```
johann-beleites-sonarsource, line 141: Can we add an explicit message about the fact that we expect no issues at all? My fear would be that someone uses the method, then there are issues raised, and they look into the sample file, assuming we called `verifyIssues()`, having a hard time figuring out that it is the expectation that is different to theirs.

johann-beleites-sonarsource (commented):
I have a couple of suggestions just for clarity of the messaging for users of the library, who may not always be super familiar with it.

test-commons/src/main/java/org/sonarsource/analyzer/commons/checks/verifier/internal/InternalIssueVerifier.java
leonardo-pilastri-sonarsource: What about "ERROR: No issues were expected, but some were found."?

johann-beleites-sonarsource (approved):

PR #332, opened 2024-05-28
[DO NOT MERGE] ACOMMONS-14: Add SonarLintCache component

anton-haubner-sonarsource:
`SonarLintCache` is a component that can be used to transfer data in memory between plugins in a SonarLint context, using the existing caching mechanisms.

At the time of writing, this component used only by the DBD plugin, to consume IRs produced by DBD custom rules.
We place it in analyzer-commons, since it is a depencency of multiple language plugins (sonar-java and sonar-python).

The component is just an intermediate solution until a dedicated mechanism to communicate between plugins with sufficient capabilities is available.


anton-haubner-sonarsource:
In the end, this PR was not merged during the current MMF-3713 sprint since the dependency injection mechanism apparently can not inject components from API packages which are not part of a plugin, and no plugin is provided by analyzer-commons.

Hence, this change is put on hold, until this limitation is resolved, or a different mechanism is introduced to transfer data between plugins.

PR #334, opened 2024-07-09
ACOMMONS-15: Add convenience factory methods for small collections

SolalPirelli-Sonar:
I added the exact same overloads that Java's `Map, Set, List` have, so up to 10 items each. I think consistency with the JDK outweighs other concerns in this case (e.g., by default SonarLint doesn't like >7 params).

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/collections/PSet.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/collections/PSet.java
@@ -65,4 +65,224 @@ public interface PSet<E> extends Iterable<E> {
   default Stream<E> stream() {
     return StreamSupport.stream(this.spliterator(), false);
   }
+
+  /**
+   * Returns a persistent set containing zero elements.
+   *
+   * @param <E> the {@code PSet}'s element type
+   * @return an empty {@code PSet}
+   */
+  static <E> PSet<E> of() {
+    return PCollections.emptySet();
+  }
+
+  /**
+   * Returns a persistent set containing one element.
+   *
+   * @param <E> the {@code PSet}'s element type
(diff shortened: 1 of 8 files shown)


commons/src/main/java/org/sonarsource/analyzer/commons/collections/PSet.java
```diff
@@ -65,4 +65,224 @@ public interface PSet<E> extends Iterable<E> {
...
+  }
+
+  /**
+   * Returns a persistent set containing one element.
+   *
+   * @param <E> the {@code PSet}'s element type
+   * @param e1 the single element
+   * @return a {@code PSet} containing the specified element
+   */
+  static <E> PSet<E> of(E e1) {
```
andreaguarino, line 86: Maybe for set and stack it makes sense to implement a single method with variadic arguments?
```java
static <E> PSet<E> of(E... elements) {
// ...
```

SolalPirelli-Sonar: That allocates an array per call, which is I believe the reason why the JDK doesn't do it. I don't know how much we care.

andreaguarino: That's a fair point; I'm not really strong on this one!

SolalPirelli-Sonar: I think "let's do whatever the JDK does" probably wins over our individual opinions here. I also find it weird that there are overloads up to 10 things, but... that's what the JDK does.

SolalPirelli-Sonar:
@andreaguarino Thanks for the review!

PR #335, opened 2024-07-26
ACOMMONS-11 Expose STIG metadata in analyzer-commons

sallaigy:
PR on the RSPEC side to update the keys in the JSON rule metadata: https://github.com/SonarSource/rspec/pull/4098/files


commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
```diff
@@ -500,6 +501,19 @@ class TestRule {
       .containsExactlyInAnyOrder("cwe:311", "cwe:315", "cwe:614", "owaspTop10:a2", "owaspTop10:a3", "owaspTop10-2021:a4", "owaspTop10-2021:a5");
   }
 
+  @Test
+  public void test_security_standards_on_10_10_return_stig() {
```
andreaguarino, line 505: should we have a test for before 10.10?

andreaguarino (approved):
LGTM! Can you please test locally with SQ to make sure it's working end-to-end?

commons/src/test/java/org/sonarsource/analyzer/commons/RuleMetadataLoaderTest.java
sallaigy: I wanted to do it, but it's already covered by the test below (`test_security_standards_on_9_9_return_asvs`). It reads the same file and checks the same expectations that we we would have had on 9.9. Or would you like to have a test for let's say 10.9?

andreaguarino: ahh I see! I think that's good then!

PR #339, opened 2024-08-16
ACOMMONS-16: Fix AVLTree iteration when buckets have multiple items

SolalPirelli-Sonar:
As I wrote in [the ticket](https://sonarsource.atlassian.net/browse/ACOMMONS-16), the `forEach` test already passed but the `iterator` one did not.

Seems to go all the way back to #240 .

### DIFF
--- a/commons/src/test/java/org/sonarsource/analyzer/commons/collections/AVLTreeTest.java
+++ b/commons/src/test/java/org/sonarsource/analyzer/commons/collections/AVLTreeTest.java
@@ -503,29 +503,73 @@ public class AVLTreeTest {
 
   @Test
   public void test_not_equal_but_same_hashcode_iteration() {
-    var a = new Weird();
-    var b = new Weird();
+    var set = PSet.of(
+      new Weird(0),
+      new Weird(0)
+    );
 
     int count = 0;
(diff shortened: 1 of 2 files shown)


commons/src/test/java/org/sonarsource/analyzer/commons/collections/AVLTreeTest.java
```diff
@@ -499,4 +500,40 @@ public void test_stack_stream() {
     Stream<Integer> stream = stack.stream();
     assertThat(stream).containsExactly(3, 2, 1);
   }
+
```
anton-haubner-sonarsource, line 503: Nitpick: How about a test for a more complex set that mixes objects with and without the same hash code?

anton-haubner-sonarsource (approved):
LGTM! Thank you for discovering and fixing this!

One nitpicky comment below.

commons/src/test/java/org/sonarsource/analyzer/commons/collections/AVLTreeTest.java
SolalPirelli-Sonar: Done.

PR #340, opened 2024-08-22
SONARPY-2080 Move ShannonEntropy to analyzer commons

ghislainpiot:


commons/src/main/java/org/sonarsource/analyzer/commons/ShannonEntropy.java
```diff
@@ -0,0 +1,44 @@
...
+    if (str == null || str.isEmpty()) {
+      return 0.0d;
+    }
+    int length = str.length();
+    return str.chars()
+      .collect(HashMap<Integer, Integer>::new, (map, ch) -> map.merge(ch, 1, Integer::sum), HashMap::putAll)
+      .values().stream()
+      .mapToDouble(count -> ((double) count) / length)
+      .map(frequency -> -frequency * Math.log(frequency))
+      .sum() / LOG_2;
```
ghislainpiot, line 42: I moved the division from the `map` to outside of it, it does not seem to affect numerical stability in the few tests

commons/src/main/java/org/sonarsource/analyzer/commons/ShannonEntropy.java
```diff
@@ -0,0 +1,44 @@
...
+    // utility class
+  }
+
+  public static double calculate(@Nullable String str) {
+    if (str == null || str.isEmpty()) {
+      return 0.0d;
+    }
+    int length = str.length();
+    return str.chars()
+      .collect(HashMap<Integer, Integer>::new, (map, ch) -> map.merge(ch, 1, Integer::sum), HashMap::putAll)
```
maksim-grebeniuk-sonarsource, line 38: could be simplifies:
```java
return str.chars()
        .boxed()
        .collect(Collectors.groupingBy(Function.identity(), Collectors.counting()))
        .values()
        .stream()
        .map(Long::doubleValue)
        .mapToDouble(count -> count / length)
        .map(frequency -> -frequency * Math.log(frequency))
        .sum() / LOG_2;
```

maksim-grebeniuk-sonarsource (approved):
LGTM, just one super minor suggestion to simplify the stream API used

commons/src/main/java/org/sonarsource/analyzer/commons/ShannonEntropy.java
ghislainpiot: I was reusing the previous SonarJava implementation, but if this one also works I can change it.

PR #344, opened 2024-10-04
SONARXML-146 Allow checks to access SensorContext to read configuration

leonardo-pilastri-sonarsource:
In the context of implementing XMLChecks that rely on analysis properties, like "sonar.android.minsdkversion.min", which are gathered by the scanners and stored into the `SensorContext.config()`, this PR wants to allow SonarXMLCheck implementations to access the context, and allow the SonarXMLCheckVerifier to pass this kind of configuration to properly test the expected behavior of such checks.


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

PR #345, opened 2024-10-04
ACOMMONS-18: Support Multi-Quality Rule (MQR) mode

Seppli11:

PR #346, opened 2024-10-07
Bump sonar-plugin-api to  and sonar-plugin-api-impl

dorian-burihabwa-sonarsource:
sonar-plugin-api -> 10.11.0.2468
sonar-plugin-api-impl -> 10.7.96327


dorian-burihabwa-sonarsource:
Covered by #345

PR #351, opened 2024-11-04
SONARPHP-1555 Move helper classes for hard-coded secrets to analyzer commons

mstachniuk:


commons/src/test/java/org/sonarsource/analyzer/commons/EntropyDetectorTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class EntropyDetectorTest {
+
+  @ParameterizedTest
+  @MethodSource("stringPerLevelForEntropyScore")
+  void test_progressive_entropy_score_sensibility(String input, int currentSensibility) {
```
GabinL21, line 35: We should use camelCase for test method names I believe

commons/src/main/java/org/sonarsource/analyzer/commons/EntropyDetector.java
```diff
@@ -0,0 +1,56 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.analyzer.commons;
+
+public class EntropyDetector {
+  private static final int DEFAULT_MIN_SECRET_LENGTH_FOR_GIVEN_ENTROPY = 25;
+  private static final double DEFAULT_ENTROPY_INCREASE_FACTOR_BY_MISSING_CHARACTER = 1.034;
+  private static final double DEFAULT_ENTROPY_SCORE_INCREMENT = 0.6;
```
GabinL21, line 25: These should be `public` I think, for example, to use in the constructor (e.g. we only want to override the `entropyScoreIncrement`)

commons/src/main/java/org/sonarsource/analyzer/commons/LatinAlphabetDetector.java
```diff
@@ -0,0 +1,151 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.analyzer.commons;
+
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class LatinAlphabetDetector {
```
GabinL21, line 26: This is more of a `HumanLanguageDetector` no? Since the public method is `humanLanguageScore`

GabinL21 (approved):
LGTM! Please check my small comments. 

I'd personally still ask for a second review from JVM since it's their code, and they will need to use these classes in Sonar-Java. For instance, are we sure that we don't want a parametrized `RandomnessDetector` in the commons that combines entropy and human language score?

commons/src/test/java/org/sonarsource/analyzer/commons/EntropyDetectorTest.java
mstachniuk: There is no such convention in this project, mostly, snake_case is used. I will change to camelCase as it is preferred by me and to be consistent with `LatinAlphabetDetectorTest`.

commons/src/main/java/org/sonarsource/analyzer/commons/EntropyDetector.java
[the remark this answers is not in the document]
mstachniuk: It looks like the recognizers project is about scanning files line by line, I don't believe that classes from this PR belong there.

mstachniuk:
> For instance, are we sure that we don't want a parametrized RandomnessDetector in the commons that combines entropy and human language score?

I'm against combining them in this project. Now the classes are independent. The analyzer-commons should contain simple building blocks, every consumer may use them together or independently.

commons/src/main/java/org/sonarsource/analyzer/commons/EntropyDetector.java
```diff
@@ -0,0 +1,56 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.analyzer.commons;
+
+public class EntropyDetector {
```
mstachniuk, line 22: Inspiration: https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/java/org/sonar/java/checks/helpers/RandomnessDetector.java#L24

commons/src/main/java/org/sonarsource/analyzer/commons/HumanLanguageDetector.java
```diff
@@ -0,0 +1,151 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarsource.analyzer.commons;
+
+import java.util.Locale;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class HumanLanguageDetector {
```
mstachniuk, line 26: Inspiration: https://github.com/SonarSource/sonar-java/blob/master/java-checks/src/main/java/org/sonar/java/checks/helpers/LatinAlphabetLanguagesHelper.java#L27

PR #387, opened 2026-04-08
CPP-7789 Add shared SonarResolve parser to analyzer commons

mostafa-mohammed-sonarsource:
This PR moves the `SonarResolve` parser to `sonar-analyzer-commons` so it can be reused by other analyzers.

The initial implementation was released in CFamily 6.79. The main reason for moving it here is that we now need to improve and share the parsing logic, especially around code-formatting cases that can split or otherwise alter comments across multiple lines and break `sonar-resolve` directives.

This PR adds the shared parser and its tests to analyzer commons so follow-up fixes and enhancements can be implemented once and consumed from a common place. It is consumed in https://github.com/SonarSource/sonar-cpp/pull/6115.

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
@@ -27,20 +27,26 @@ public final class SonarResolve {
 
   public static final String KEYWORD = "sonar-resolve";
 
-  private final int line;
+  private final int directiveLine;
+  private final int targetLine;
   private final IssueResolution.Status status;
   private final Set<RuleKey> ruleKeys;
   private final String justification;
 
-  public SonarResolve(int line, IssueResolution.Status status, Set<RuleKey> ruleKeys, String justification) {
-    this.line = line;
+  public SonarResolve(int directiveLine, int targetLine, IssueResolution.Status status, Set<RuleKey> ruleKeys, String justification) {
+    this.directiveLine = directiveLine;
+    this.targetLine = targetLine;
     this.status = Objects.requireNonNull(status);
     this.ruleKeys = Collections.unmodifiableSet(new LinkedHashSet<>(ruleKeys));
     this.justification = Objects.requireNonNull(justification);
   }
 
-  public int line() {
-    return line;
+  public int directiveLine() {
+    return directiveLine;
+  }
+
+  public int targetLine() {
+    return targetLine;
   }
 
   public IssueResolution.Status status() {
@@ -64,7 +70,8 @@ public final class SonarResolve {
       return false;
     }
     SonarResolve other = (SonarResolve) object;
-    return line == other.line
+    return directiveLine == other.directiveLine
+      && targetLine == other.targetLine
       && status == other.status
       && ruleKeys.equals(other.ruleKeys)
       && justification.equals(other.justification);
@@ -72,13 +79,14 @@ public final class SonarResolve {
 
   @Override
   public int hashCode() {
-    return Objects.hash(line, status, ruleKeys, justification);
+    return Objects.hash(directiveLine, targetLine, status, ruleKeys, justification);
   }
 
   @Override
   public String toString() {
     return "SonarResolve{"
-      + "line=" + line
+      + "directiveLine=" + directiveLine
+      + ", targetLine=" + targetLine
       + ", status=" + status
       + ", ruleKeys=" + ruleKeys
       + ", justification='" + justification + '\''
@@ -96,19 +104,21 @@ public final class SonarResolve {
     private static final String PREFIX = "Invalid sonar-resolve directive: ";
 
     private final StringBuilder accumulatedDirective = new StringBuilder();
-    private final int line;
+    private final int directiveLine;
     private State state = State.INCOMPLETE;
     private SonarResolve result;
     private String errorMessage;
 
-    public Driver(int line) {
-      this.line = line;
+    public Driver(int directiveLine) {
+      this.directiveLine = directiveLine;
     }
 
-    public State consumeLine(String normalizedLine) {
+    // lineNumber is reserved for future features such as supporting relative offsets for target-line computation.
+    // The parameter is part of the API already so analyzers can adopt those features later without another signature change.
+    public State consumeLine(int lineNumber, String lineContent) {
       checkState(state == State.INCOMPLETE, "Cannot consume additional lines after parser reached a terminal state.");
-      appendNormalizedLine(normalizedLine);
-      Parser parser = new Parser(line);
+      appendNormalizedLine(lineContent);
+      Parser parser = new Parser(directiveLine);
       update(parser, parser.parse(accumulatedDirective.toString()));
       return state;
     }
@@ -157,15 +167,15 @@ public final class SonarResolve {
 
     private static final class Parser {
 
-      private final int line;
+      private final int directiveLine;
       private State state = State.INCOMPLETE;
       private String errorMessage;
       private IssueResolution.Status status = IssueResolution.Status.DEFAULT;
       private final Set<RuleKey> ruleKeys = new LinkedHashSet<>();
       private String justification;
 
-      private Parser(int line) {
-        this.line = line;
+      private Parser(int directiveLine) {
+        this.directiveLine = directiveLine;
       }
 
       private State parse(String accumulatedDirective) {
@@ -200,7 +210,7 @@ public final class SonarResolve {
         if (state != State.COMPLETE) {
           return null;
         }
-        return new SonarResolve(line, status, ruleKeys, justification);
+        return new SonarResolve(directiveLine, directiveLine, status, ruleKeys, justification);
       }
 
       private String errorMessage() {
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
@@ -93,7 +93,7 @@ public final class SonarResolve {
       + '}';
   }
 
-  public static final class Driver {
+  public static final class StreamingParser {
 
     public enum State {
       INCOMPLETE,
@@ -109,7 +109,7 @@ public final class SonarResolve {
     private SonarResolve result;
     private String errorMessage;
 
-    public Driver(int directiveLine) {
+    public StreamingParser(int directiveLine) {
       this.directiveLine = directiveLine;
     }
 
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
@@ -1,0 +1,402 @@
+/*
+ * SonarSource Analyzers Commons
+ * Copyright (C) 2009-2025 SonarSource Sàrl
+ * mailto:info AT sonarsource DOT com
+ *
+ * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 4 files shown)


mostafa-mohammed-sonarsource:
This change is part of the following stack:

- #387 ◀
    - #388
        - #389

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
```diff
@@ -0,0 +1,412 @@
...
+package org.sonarsource.analyzer.commons;
+
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Objects;
+import java.util.Set;
+import org.sonar.api.batch.sensor.issue.IssueResolution;
+import org.sonar.api.rule.RuleKey;
+
+public final class SonarResolve {
```
lijun-chen-sonarsource, line 26: Can we make `SonarResolve` a `record`? To save the boilerplate.

lijun-chen-sonarsource: Never mind. Saw this
https://github.com/SonarSource/sonar-analyzer-commons/blob/8751a255168c0d8dbc9da20c1264221012886c88/pom.xml#L88

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
```diff
@@ -0,0 +1,412 @@
...
+    return "SonarResolve{"
+      + "directiveLine=" + directiveLine
+      + ", targetLine=" + targetLine
+      + ", status=" + status
+      + ", ruleKeys=" + ruleKeys
+      + ", justification='" + justification + '\''
+      + '}';
+  }
+
+  public static final class Driver {
```
lijun-chen-sonarsource: The name `Driver` is slightly confusing when first approaching the code. The following code in the test confirms that `Driver` is really a parser.

https://github.com/SonarSource/sonar-analyzer-commons/blob/8751a255168c0d8dbc9da20c1264221012886c88/commons/src/test/java/org/sonarsource/analyzer/commons/SonarResolveTest.java#L43

Though I agree the current separation between `Driver` and `Parser` is meaningful so wouldn't know how to name `Driver` better. One possibility (I'm not sure I like it...): `Driver` -> `Parser`, `Parser` -> `ParserCore`.

But I'm also fine with `Driver`.

WDYT?

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
```diff
@@ -0,0 +1,412 @@
...
+    private SonarResolve result;
+    private String errorMessage;
+
+    public Driver(int directiveLine) {
+      this.directiveLine = directiveLine;
+    }
+
+    // lineNumber is reserved for future features such as supporting relative offsets for target-line computation.
+    // The parameter is part of the API already so analyzers can adopt those features later without another signature change.
+    public State consumeLine(int lineNumber, String lineContent) {
```
lijun-chen-sonarsource, line 118: Is `lineNumber` really necessary? Could we just keep an counter internally in the class that is increemnted every time `consumeLine` is called?

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
```diff
@@ -0,0 +1,412 @@
...
+      checkState(state == State.INCOMPLETE, "Cannot consume additional lines after parser reached a terminal state.");
+      appendNormalizedLine(lineContent);
+      Parser parser = new Parser(directiveLine);
+      update(parser, parser.parse(accumulatedDirective.toString()));
```
lijun-chen-sonarsource, line 122: Reparsing each time feels theoretically wasteful, since it's overall O(n^2) to the number of lines. Though in real-world I'd expect n to be < 10 in most cases.
                                                                      
I am curious whether you have considered to implement `Parser` entirely as a push-based parser to which we can feed characters. This would require implementing it as state machine, instead of the current recursive-descent approach that keeps state implicitly in control-flow & call stack and just pulls in characters when necessary. Too bad Java doesn't support generators...
                                                                      
Also, IIUC, the reason `Driver` exists is exactly to expose a push-based parser API when the underlying `Parser` is pull-based.

Just out of curiosity. Not asking any change, as most probably wouldn't be worth it and we can change the implementation in a backward compatible manner in the future anyway, if performance issues really start to arise.

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
```diff
@@ -0,0 +1,412 @@
...
+            return '\t';
+          default:
+            return character;
+        }
+      }
+
+      private boolean incomplete(String message) {
+        state = State.INCOMPLETE;
+        errorMessage = message;
+        justification = null;
```
lijun-chen-sonarsource, line 329: Nitpick:
This is surely `null` at this point right? IIUC, `justification != null` <=> `state == State.COMPLETE` in the current code.
But I can see also a pro. Setting all the parser outputs is nice for consistency.
Feel free to ignore.

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
```diff
@@ -0,0 +1,412 @@
...
+        return false;
+      }
+
+      private boolean invalid(String message) {
+        state = State.INVALID;
+        errorMessage = message;
+        justification = null;
```
lijun-chen-sonarsource, line 336: Same as above.

commons/src/test/java/org/sonarsource/analyzer/commons/SonarResolveTest.java
```diff
@@ -0,0 +1,294 @@
...
+        new SonarResolve(42, 42, IssueResolution.Status.DEFAULT, Set.of(RuleKey.of("cpp", "S100")), "line \"comment\"")),
+      arguments(
+        "sonar-resolve cpp:S100 \"line\\nline\\rline\\tline\\\\line\"",
+        42,
+        new SonarResolve(
+          42,
+          42,
+          IssueResolution.Status.DEFAULT,
+          Set.of(RuleKey.of("cpp", "S100")),
+          "line\nline\rline\tline\\line")));
```
lijun-chen-sonarsource, line 217: How about adding a test like `"sonar-resolve cpp:S100 \"line comment\" the rest is ignored"`.
It shows that after the justification finishes, the parser ignores the rest of the line and doesn't go further in the input.

commons/src/test/java/org/sonarsource/analyzer/commons/SonarResolveTest.java
```diff
@@ -0,0 +1,294 @@
...
+          "middle\\t",
+          "suffix\\\"\""
+        },
+        42,
+        new SonarResolve(
+          42,
+          42,
+          IssueResolution.Status.DEFAULT,
+          Set.of(RuleKey.of("cpp", "S1234")),
+          "prefix\n\nmiddle\t\nsuffix\"")));
```
lijun-chen-sonarsource: How about adding a test like
                                                                                                                                                
```java
new String[] {
   "sonar-resolve",
   "cpp:S1234 \"reason\" the rest is ignored"
}
```
                                                                                                                                                
Showing that after the justification finishes, the parser ignores the rest of the line?
                                                                                                                                                
Also, do you expect analyzers to make use of the portion of the comment that follow a complete sonar-resolve directive?
E.g., in the example, " the rest is ignored" is still in a line that is partially covered by a sonar-resolve directive.
Do you think analyzers may want to know where (i.e. which column) the `sonar-resolve` directive exactly ended? Instead of just at which line it completed?

lijun-chen-sonarsource (changes requested):
Thanks for working on this. I personally very much like `sonar-resolve` as feature, so looking forward to see everything merged and released.

I left some comments. Curious to hear what you think.

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
lijun-chen-sonarsource: I guess it depends on when offsets will be finalized whether we'll want to support something like:

```cpp
// sonar-resolve + cpp:S100 "reason 

// applies to code2"
```

I.e., non-contiguous comment lines that form a sonar-resolve directive.

This is one case I can think of to require `lineNumber`. I'd curious to see other use-cases.

Edit: from `sonar-cpp/sonar-cfamily-plugin/src/test/resources/sonar-resolve/invalid-single-line-multiline.cpp` I can see that we treat this as invalid.

mostafa-mohammed-sonarsource: I agree we could count lines internally with the current behavior. The reason I kept it in the API is future-proofing for offset-based target selection: if implicit counting turns out not to be sufficient once that feature is specified, we would have to change the API again, which defeats the purpose of preparing it now.

regarding the empty-lines/gaps: the main reason we support continuation lines is to be resilient to code formatters such as clang-format: they may reflow a long comment onto subsequent lines, but they should not introduce empty gaps between those lines. So this is intended both as a strict extension for formatter-resilience and as a small ergonomic improvement for users writing longer justifications.

So what do you think of keeping it?

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
mostafa-mohammed-sonarsource: Renamed to StreamingParser

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
mostafa-mohammed-sonarsource: The current state evolved naturally with Parser first, because sonar-resolve was limited to a single line, so I agree it's kind of wasteful, but we can revisit it later if it proves to be a problem

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
mostafa-mohammed-sonarsource: You're correct, I did it as a form of explicitly setting the full state instead of being implicit

commons/src/test/java/org/sonarsource/analyzer/commons/SonarResolveTest.java
mostafa-mohammed-sonarsource: Added, thank you 🙇‍♂️

commons/src/test/java/org/sonarsource/analyzer/commons/SonarResolveTest.java
mostafa-mohammed-sonarsource: I don’t currently expect analyzers to need the trailing portion after a complete sonar-resolve directive.

If we later discover a real analyzer need to preserve or further interpret the remainder of the comment, then exposing an end column or consumed range would be a reasonable extension. For now, though, that feels premature.

commons/src/main/java/org/sonarsource/analyzer/commons/SonarResolve.java
lijun-chen-sonarsource: Yeah. Makes sense to me. Let's keep it. If we really find out that even with offset-based target selection in 99% of the cases counting lines internally is sufficient, we can later anyway add an overload to `consumeLine` that doesn't require `linuNumer`. And it would be backward compatible.

lijun-chen-sonarsource (approved):
LGTM!

PR #395, opened 2026-06-05
APPSEC-3359: Add CleartextProtocolFilter for S5332 safe-URL detection

daniel-teuchert-sonarsource:
## Summary

Adds `CleartextProtocolFilter.isSafeWithoutTls(String url)` to the `commons` module as a shared oracle for S5332 (cleartext protocol) — and potentially other rules in the future.

Two FP categories are covered, identified via analysis of ~4 800 SAFE-marked hotspots:

- **Internal / non-public hosts** — loopback (IPv4/IPv6), full IPv4 link-local range `169.254.0.0/16` (covers AWS, Azure, GCP, OCI IMDS), AWS IPv6 IMDS, Azure wireserver (`168.63.129.16`), Alibaba Cloud ECS IMDS (`100.100.100.200`), GCP/generic IMDS hostnames, Docker-internal, Kubernetes `*.svc.cluster.local`. None reachable from the public internet.
- **Namespace URI authorities** — W3C, Android, Microsoft, OASIS, HL7, Spring, Maven, and more. These `http://` URIs are opaque identifiers mandated by standards; no actual HTTP connection is implied.
- **IANA-reserved documentation domains** — `example.com/net/org` and their subdomains, plus RFC 6761 reserved TLDs `.example`, `.test`, `.localhost`. Almost always placeholders in source code.

Notable design decisions:
- `0.0.0.0` and ASP.NET `+` (listen wildcards) were **explicitly excluded** — they bind to all interfaces including public-facing ones and should trigger S5332.
- The `url` parameter is documented non-null (`@ParametersAreNonnullByDefault` applies via `package-info.java`).

Each language analyzer currently carries its own inline loopback exemption; this is the intended single source of truth to consolidate them (APPSEC-3185).

## Test plan

- [x] `CleartextProtocolFilterTest` passes (JUnit 5 parameterized — 73 safe URLs, 17 not-safe URLs, all green)
- [ ] Review host / namespace-authority lists — additions and removals welcome
- [ ] Confirm class placement (`org.sonarsource.analyzer.commons` root package) is acceptable

## Follow-up

Once released, each analyzer adopts this via a version bump and replaces its own inline logic:
- `sonar-iac-enterprise`: `UrlUtils.COMPLIANT_HTTP_DOMAINS` → `CleartextProtocolFilter.isSafeWithoutTls()`
- `sonar-python-enterprise`: `ClearTextProtocolsCheck` loopback regex
- `sonar-go-enterprise`: `ClearTextProtocolCheck.isLoopback()`
- `SonarJS`: no current exemption — gains full coverage

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


commons/src/main/java/org/sonarsource/analyzer/commons/CleartextProtocolFilter.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  /**
+   * Returns {@code true} if {@code url} is safe and should NOT trigger a
+   * cleartext-protocol security warning. Call this before raising an issue.
+   *
+   * @param url the raw URL string as it appears in source code
+   */
+  public static boolean isSafe(String url) {
+    var matcher = CLEARTEXT_URL.matcher(url.strip());
```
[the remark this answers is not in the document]
daniel-teuchert-sonarsource: Check the package-info. It has `ParametersAreNonnullByDefault` so the url parameter can not be null.
I will add a comment for this.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
```diff
@@ -0,0 +1,191 @@
+/*
```
guillaume-dequenne, line 1: I'm not sure how many of these helpers will eventually land here, but if we plan to have more, I'm thinking we could have a `security` (or any better name) subpackage to organize this a bit more.

commons/src/main/java/org/sonarsource/analyzer/commons/CleartextProtocolFilter.java
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Returns {@code true} if {@code url} is safe to use without TLS and should NOT
+   * trigger a cleartext-protocol security warning. Call this before raising an issue.
+   *
+   * @param url the parsed URI; must not be null
+   */
+  public static boolean isSafeWithoutTls(URI url) {
+    var scheme = url.getScheme();
+    if (!scheme.equalsIgnoreCase("http") && !scheme.equalsIgnoreCase("ftp")) {
```
guillaume-dequenne: Can't `scheme` be `null` here? We probably should avoid the risk of NPE in this method.

guillaume-dequenne (changes requested):
Looks good overall, just a small comment on the risk of NPE and a broader one on whether we should have a dedicated package for this.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: Agreed! We will think about how to handle such things in the future because we already saw other opportunities for something like this. Maybe we will create a different repo for it, but for now I am moving this to an appsec subpackage.

commons/src/main/java/org/sonarsource/analyzer/commons/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: whoopsie. Thanks for spotting this!

daniel-teuchert-sonarsource:
> Looks good overall, just a small comment on the risk of NPE and a broader one on whether we should have a dedicated package for this.

Thanks for your review. I applied your feedback and while doing that noticed that we can do some improvements to the scheme filtering. I added more clear text filters and this caused the JAR to be too large. I adjusted the accepted size in the pom.xml file.
Please review if that is okay. It was just a few bytes above the acceptable limit.

commons/pom.xml
```diff
@@ -122,7 +122,7 @@
                   <!-- Warning: Please consider carefully when increasing the size of this shared library, it might have
                   impact on all our analyzers! -->
                   <minsize>140000</minsize>
-                  <maxsize>150000</maxsize>
+                  <maxsize>151000</maxsize>
```
guillaume-dequenne: You probably can increase to `160000` here to avoid having to redo this as soon as we make other changes. That kind of increase should be quite okay.

guillaume-dequenne (approved):
LGTM, left one minor suggestion.

commons/pom.xml
daniel-teuchert-sonarsource: Will do. Thanks!

PR #398, opened 2026-06-11
APPSEC-3389: Expose getCleartextProtocols() and getAlternativeProtocols() in CleartextProtocolFilter

daniel-teuchert-sonarsource:
## Summary

- Adds `getCleartextProtocols()` returning the set of cleartext scheme prefixes with `://` (e.g. `{"http://", "ftp://", ...}`) for use in startsWith-style detection
- Adds `getAlternativeProtocols()` returning a map of scheme name → secure alternative (e.g. `{"http" → "https", ...}`) for use in issue messages
- `CLEARTEXT_PROTOCOL_ALTERNATIVES` is now the single source of truth; `CLEARTEXT_SCHEMES` is derived from its keySet to eliminate duplication

## Test plan

- [ ] `getCleartextProtocolsReturnsExpectedSet` — verifies all 17 scheme prefixes are present
- [ ] `getCleartextProtocolsIsUnmodifiable` — returned set rejects mutation
- [ ] `getAlternativeProtocolsReturnsExpectedMappings` — verifies all 17 scheme → alternative entries
- [ ] `getAlternativeProtocolsIsUnmodifiable` — returned map rejects mutation
- [ ] `cleartextProtocolKeysMatchAlternativeProtocolKeys` — consistency check between the two accessors

----
## Summary by Gitar

- **Refactored protocol data:**
  - Replaced the hardcoded `CLEARTEXT_SCHEMES` set with `CLEARTEXT_PROTOCOL_ALTERNATIVES` map to act as the single source of truth.
  - Added `getIssueMessage(String scheme)` to dynamically generate standardized security warnings.
- **Improved internal logic:**
  - Updated `isSafeWithoutTls` to leverage the new case-insensitive map lookups.
  - Derived `CLEARTEXT_SCHEME_PREFIXES` automatically from the map keys to ensure consistency.
- **Testing enhancements:**
  - Added comprehensive tests for `getIssueMessage` and ensured immutability of exposed collections.
  - Verified all supported protocols have associated security messages.

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
@@ -19,8 +19,10 @@ package org.sonarsource.analyzer.commons.appsec;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * Determines whether a cleartext-protocol URL should be considered safe and should not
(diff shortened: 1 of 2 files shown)


commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
```diff
@@ -128,9 +130,35 @@ public final class CleartextProtocolFilter {
...
 
-  private static final Set<String> CLEARTEXT_SCHEMES = Set.of(
-    "http", "ftp", "ws", "telnet", "rtmp", "tftp", "gopher", "irc",
-    "smtp", "ldap", "amqp", "mqtt", "imap", "pop3", "nntp", "sip", "stomp");
+  /**
+   * Maps each cleartext scheme name to its recommended secure alternative.
+   * This is the single source of truth for both detection and messaging.
+   */
+  private static final Map<String, String> CLEARTEXT_PROTOCOL_ALTERNATIVES = Map.ofEntries(
+    Map.entry("http",    "https"),
```
teemu-rytilahti-sonarsource: Would it make sense to use records here to allow storing some extra data (like user readable string) instead? I was looking into fixing an ugly report string (missing/inconsistent period, lowercase protocol acronym) for this rule on sonar-python, and this would cover that case, too.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
```diff
@@ -128,9 +130,35 @@ public final class CleartextProtocolFilter {
...
+    Map.entry("ldap",    "ldaps"),
+    Map.entry("imap",    "imaps"),
+    Map.entry("pop3",    "pop3s"),
+    Map.entry("amqp",    "amqps"),
+    Map.entry("mqtt",    "mqtts"),
+    Map.entry("sip",     "sips"),
+    Map.entry("rtmp",    "rtmps"),
+    Map.entry("irc",     "ircs"),
+    Map.entry("nntp",    "nntps"),
+    Map.entry("stomp",   "stomps")
```
teemu-rytilahti-sonarsource: stomp?

daniel-teuchert-sonarsource: Do you think this is too obscure to search for?

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: Do you already have any extra data to add here in mind? If not, should I still already change this to records?

teemu-rytilahti-sonarsource: Oh, this is already for the user readable part, not just to get a secure replacement scheme for an insecure variant... Sorry for the confusion.

As discussed offline, maybe it makes rather sense to add a `getSecureAlternativeMessage(proto)` to make the UI strings consistent across the analyzers?

One thing to keep in mind if doing that is to make sure there is some sort of default fallback if the protocol mapping does not exist.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
teemu-rytilahti-sonarsource: It sounds a bit obscure, but I doubt it does any harm to have it.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: Agreed!

teemu-rytilahti-sonarsource (approved):

PR #399, opened 2026-06-11
APPSEC-3388: Add missing namespace URI authorities, fix bare-scheme bug, widen IPv4 loopback

daniel-teuchert-sonarsource:
## Summary

- Adds 10 missing namespace URI authorities to `NAMESPACE_URI_AUTHORITIES`: `xml.apache.org`, `schemas.openxmlformats.org`, `rdfs.org`, `schemas.google.com`, `a9.com`, `ns.adobe.com`, `ltsc.ieee.org`, `docbook.org`, `graphml.graphdrawing.org`, `json-schema.org`
- Fixes bare-scheme bug: `http://` was incorrectly returned as safe in the lenient fallback path when `CLEARTEXT_AUTHORITY` found no authority; a new `CLEARTEXT_SCHEME_PREFIX` pattern is used to distinguish "not a cleartext URL" from "cleartext URL with no host"
- Widens IPv4 loopback pattern from `{3}` to `{1,3}` to accept short-form addresses (`127.1`, `127.0.1`) that resolve to loopback on POSIX systems

## Test plan

- [ ] 10 new safe URL test cases for each added namespace URI authority
- [ ] `http://`, `ftp://`, `ws://` added to `notSafeUrls` — verify bare-scheme bug is fixed
- [ ] `http://127.1`, `http://127.0.1`, `http://127.1:8080` added to `safeUrls` — verify short-form loopback

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
@@ -196,11 +196,6 @@ public final class CleartextProtocolFilter {
   private static final Pattern CLEARTEXT_AUTHORITY = Pattern.compile(
     "^(?:" + String.join("|", CLEARTEXT_SCHEMES) + ")://(?:[^@\\s/?#]++@)?(?<rest>[^\\s/?#]++)", Pattern.CASE_INSENSITIVE);
 
-  // Matches a bare cleartext scheme prefix (e.g. "http://") with no authority following it.
-  // Used to detect bare-scheme strings that CLEARTEXT_AUTHORITY won't match.
-  private static final Pattern CLEARTEXT_SCHEME_PREFIX = Pattern.compile(
-    "^(?:" + String.join("|", CLEARTEXT_SCHEMES) + ")://", Pattern.CASE_INSENSITIVE);
-
   private CleartextProtocolFilter() {
   }
 
@@ -253,7 +248,7 @@ public final class CleartextProtocolFilter {
       // fall through to lenient parsing
     }
     var matcher = CLEARTEXT_AUTHORITY.matcher(stripped);
-    return matcher.find() ? isSafeHost(matcher.group("rest")) : !CLEARTEXT_SCHEME_PREFIX.matcher(stripped).find();
+    return matcher.find() ? isSafeHost(matcher.group("rest")) : CLEARTEXT_SCHEME_PREFIXES.stream().noneMatch(p -> stripped.toLowerCase(Locale.ROOT).startsWith(p));
   }
 
   /**
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
@@ -64,8 +64,8 @@ public final class CleartextProtocolFilter {
   private static final Pattern SAFE_HOSTS = Pattern.compile("(?:" +
     // localhost
     "^localhost|" +
-    // IPv4 loopback (127.0.0.0/8)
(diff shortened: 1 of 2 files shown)


commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
```diff
@@ -59,8 +59,8 @@ public final class CleartextProtocolFilter {
   private static final Pattern SAFE_HOSTS = Pattern.compile("(?:" +
     // localhost
     "^localhost|" +
-    // IPv4 loopback (127.0.0.0/8)
-    "^127(?:\\.\\d+){3}|" +
+    // IPv4 loopback (127.0.0.0/8) — 2–4 octet forms (127.1, 127.0.1, 127.0.0.1)
```
teemu-rytilahti-sonarsource, line 67: Do we have some data to back this up? I am a bit hesitant about this as it is just an implementation gimmick. We can give it a try though!

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
```diff
@@ -166,6 +186,11 @@ public final class CleartextProtocolFilter {
   private static final Pattern CLEARTEXT_AUTHORITY = Pattern.compile(
     "^(?:" + String.join("|", CLEARTEXT_SCHEMES) + ")://(?:[^@\\s/?#]++@)?(?<rest>[^\\s/?#]++)", Pattern.CASE_INSENSITIVE);
 
+  // Matches a bare cleartext scheme prefix (e.g. "http://") with no authority following it.
+  // Used to detect bare-scheme strings that CLEARTEXT_AUTHORITY won't match.
+  private static final Pattern CLEARTEXT_SCHEME_PREFIX = Pattern.compile(
+    "^(?:" + String.join("|", CLEARTEXT_SCHEMES) + ")://", Pattern.CASE_INSENSITIVE);
```
teemu-rytilahti-sonarsource: Instead of using a regex, maybe we could have a list of a static set based on the CLEARTEXT_PROTOCOL_ALTERNATIVES keys?

teemu-rytilahti-sonarsource (approved):
LGTM, added a couple of potential ideas but feel free to proceed as you wish.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: This is what I did to confirm it works:
```
 ping 127.1
PING 127.1 (127.0.0.1): 56 data bytes
64 bytes from 127.0.0.1: icmp_seq=0 ttl=64 time=0.283 ms
64 bytes from 127.0.0.1: icmp_seq=1 ttl=64 time=0.487 ms
```
Chrome also seems to automatically convert this to 127.0.0.1.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: Good idea!

daniel-teuchert-sonarsource: agreed

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
teemu-rytilahti-sonarsource: Yes, it is done on linux by some network stack apis, but I am wondering if there is enough real-world that we should check for it in this lib. But let's keep it!

PR #394, opened 2026-06-01
Allow dashes in `sonar-resolve` rule keys

cirras:
This PR enables the `sonar-resolve` parser to recognize rule keys containing dashes.

The existing implementation only allows letters, digits, colons, and underscores. This blocks plugins like SonarDelphi (which has repository key `community-delphi`) from implementing support for `sonar-resolve`.

See: https://github.com/integrated-application-development/sonar-delphi/issues/432.

PR #404, opened 2026-06-19
ACOMMONS-58 Allow dashes in `sonar-resolve` rule keys

vdiez:
This PR cherry-picks commit `2443a9beba52ffc184d34ffd08e6000d1aeb970e` from #394 so CI can run from an in-repo branch.

The `sonar-resolve` parser should recognize rule keys containing dashes.

The existing implementation only allows letters, digits, colons, and underscores. This blocks plugins like SonarDelphi, which has repository key `community-delphi`, from implementing support for `sonar-resolve`.

See:
- #394
- integrated-application-development/sonar-delphi#432


mostafa-mohammed-sonarsource (approved):
✅

PR #422, opened 2026-07-03
APPSEC-3475 SecretClassifier: Add double-underscore-wrapped placeholder pattern

teemu-rytilahti-sonarsource:
Adds `^__.+__$` to the PLACEHOLDER pattern group in `SecretClassifier` to recognize double-underscore-wrapped values like `__some_placeholder_password__` as non-secrets. This pattern is commonly used as a placeholder convention in config files.

Noticed during review of SONARXML-374.

Part of APPSEC-3464

----
## Summary by Gitar

- **Code improvements:**
  - Added internal `classify(String)` helper to `SecretClassifier` to expose category detection logic for testing.
- **Testing:**
  - Restructured `SecretClassifierTest` to use granular `MethodSource` and `ParameterizedTest` per classification category.
  - Reorganized `KNOWN_NON_SECRETS` into categorized lists (`FAKE_VALUE_SAMPLES`, `SECRET_SAMPLES`, etc.) for better maintainability.

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
@@ -152,7 +152,9 @@ public final class SecretClassifier {
       // Python format string placeholders, e.g. "%(password)s"
       "^%\\([^)]++\\)s$",
       // Azure Logic Apps runtime expressions, e.g. "@variables('name')", "@body('action')"
-      "^@\\w++\\([^)]*+\\)$"),
+      "^@\\w++\\([^)]*+\\)$",
+      // Double-underscore-wrapped placeholders, e.g. "__some_placeholder_password__"
+      "^__.+__$"),
 
     // Encrypted markers wrapping a ciphertext.
     PatternGroup.of(Category.ENCRYPTED,
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
@@ -108,9 +108,7 @@ public final class SecretClassifier {
       // Same character repeated from start to end, e.g. "aa", "111111"
       "^(?<repeated>.)\\k<repeated>*+$",
       // A secret being masked or shortened, e.g. "1fj28...askn3i"
-      "\\.\\.\\.",
-      // Code-reminder placeholder markers left as a value (see the regex)
-      "^(?:todo|fixme)\\b"),
+      "\\.\\.\\."),
 
     // Templating, interpolation and env/config lookups where the value comes from elsewhere.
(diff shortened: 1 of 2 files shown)


commons/src/test/java/org/sonarsource/analyzer/commons/appsec/SecretClassifierTest.java
```diff
@@ -52,6 +52,7 @@ class SecretClassifierTest {
     "hunter2", "letmein", "abc123",
     "changeme", "changeit", "unknown", "optional", "enabled", "disabled", "string", "random", "token",
     // PLACEHOLDER
+    "__some_placeholder_password__",
```
daniel-teuchert-sonarsource: Can you find an example that will not be filtered out by the other filters?
E.g. `pass` already prevents this from raising.

daniel-teuchert-sonarsource (approved):
LGTM!
Just a minor suggestion for a test improvement.

In general we could think about changing tests to verify that the correct group matches (e.g. make sure that `__some_placeholder_password__` is filtered out by the placeholder group).

commons/src/test/java/org/sonarsource/analyzer/commons/appsec/SecretClassifierTest.java
teemu-rytilahti-sonarsource: Done, the newest commit adds category-based tests to make sure we have proper test coverage across different categories even if some of the filters would be modified/removed/adjusted in the future.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
```diff
@@ -255,6 +257,25 @@ static Set<String> exactMatchValues() {
     return SECRET_VALUES.values();
   }
 
+  /** Visible for testing: returns the {@link Category} that suppressed the candidate, or {@code null} when not a known non-secret. */
+  @Nullable
+  static Category classify(@Nullable String candidate) {
```
daniel-teuchert-sonarsource, line 262: There is probably no nice way to have this implementation in the test files as it is currently only used by tests?

As it might become useful in the future also for the actual behavior this is probably no issue.

daniel-teuchert-sonarsource (approved):
LGTM!
Thanks for the test improvements!

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
teemu-rytilahti-sonarsource: It is a package-private, so I think it is fine. The other alternatives are exposing some public API / information about the match result, or making it completely private & leveraging reflections to access it privately. Both of them are worse in my opinion, so I would say let's keep it like this.

daniel-teuchert-sonarsource: Yep! Both of them are worse. This is better 👍

PR #426, opened 2026-07-06
ACOMMONS-78 Expose the regexes from SecretClassifier as JSON file

mstachniuk:
Part of 

----
## Summary by Gitar

- **New module:**
  - Added `sonar-analyzer-commons-configurations` to export `SecretClassifier` patterns as a machine-readable JSON file.
  - Implemented `SecretPatternsExporter` to convert Java regexes to .NET-compatible atomic groups.
- **Packaging:**
  - Configured build to generate and bundle the JSON patterns into both `.nupkg` (NuGet) and `.tar.gz` (npm) formats.
  - Updated root `pom.xml` to include the new module and publish the resulting JSON, nupkg, and tgz artifacts.
- **API changes:**
  - Added `exportPatternGroups()` and `exportExactMatchGroups()` to `SecretClassifier` for external consumption of classification logic.

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
@@ -21,7 +21,6 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /**
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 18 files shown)


sonar-analyzer-commons-configurations/src/main/packaging/nuget/sonar-analyzer-commons-configurations.nuspec
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd">
+  <metadata>
+    <id>SonarAnalyzer.CommonsConfigurations</id>
+    <version>${package.version}</version>
+    <authors>SonarSource</authors>
+    <owners>SonarSource</owners>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <license type="expression">LGPL-3.0-only</license>
```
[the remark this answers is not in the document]
mstachniuk: The whole project is LGPL-3.0 so it is ok. It should be fixed in separate PR

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
```diff
@@ -275,6 +275,81 @@ static Category classify(@Nullable String candidate) {
...
+   * Returns the skip-pattern groups as raw regex source strings, grouped by {@link Category}, in declaration order.
+   *
+   * <p>Exposed so the build can emit a single machine-readable export (JSON) of the very patterns the JVM classifier
+   * uses, keeping non-JVM analyzers (SonarJS, sonar-dotnet, …) in sync without duplicating the list. The patterns are
+   * applied case-insensitively with "find" (search-anywhere) semantics, matching {@link #isKnownNonSecret(String, Context)}.
+   *
+   * @return an immutable, ordered list of pattern groups
+   */
+  public static List<PatternGroupView> exportPatternGroups() {
+    return PATTERN_GROUPS.stream()
```
maksim-grebeniuk-sonarsource, line 288: minor, I'm not sure how much times do we use this method, maybe it makes sense to cache the result in a field?

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
```diff
@@ -275,6 +275,81 @@ static Category classify(@Nullable String candidate) {
...
+  }
+
+  /**
+   * Returns the exact-match value groups, matched in full and case-insensitively, grouped by {@link Category}.
+   * Values are sorted so the export is deterministic.
+   *
+   * @return an immutable, ordered list of exact-match groups
+   */
+  public static List<ExactMatchGroupView> exportExactMatchGroups() {
+    return List.of(new ExactMatchGroupView(
```
maksim-grebeniuk-sonarsource, line 302: minor, I'm not sure how much times do we use this method, maybe it makes sense to cache the result in a field?

maksim-grebeniuk-sonarsource (approved):
LGTM. see my small comment

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
mstachniuk: It is called only once during the build time, there is no sense to cache it

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
mstachniuk: It is called only once during the build time, there is no sense to cache it

Issue #212, opened 2022-02-07
The regex parser silently parses the malformed character class '[\\\\]'


stanislav-jordanov-sonarsource:
1. `'[\]'` -> CharClass("\")  (?)
2. `'[\]a'` -> CharClass("\") ... (??) 
3. `'[\]a]'` -> CharClass("]a") (OK)
4. `'[\\]'` -> CharClass("\")  (OK)
5. `'[\\\]'` -> CharClass("\")  (???)
6. `'[\\\a'` -> CharClass("\")   (???)
7. `'[\\\a]'` -> CharClass("\a")  (OK)
8. `'[\\\]a'` -> CharClass("\") ... (??)
9. `'[\\\]a]'` -> CharClass("\") ... (???)
10. `'[\\\\]'` -> CharClass("\") ... (OK-ish but duplicate backslash is lost in parsing)

The above list shows how different regex string literals are parsed. Well not parsed-parsed, but only the first element of the parsing - always a CharacterClass in this case, but in a lot of cases a pretty surprising one.

Wohops:
Imported to Jira - [ACOMMONS-3](https://sonarsource.atlassian.net/browse/ACOMMONS-3)

PR #431, opened 2026-07-13
ACOMMONS-3 Add regressions tests

mallory-scotton:
## Summary

ACOMMONS-3 (imported from #212) says the regex parser silently turns malformed character class literals into unexpected ASTs. The original report is a list of 10 backslash-heavy strings, each with a `?`/`??`/`???` next to the ones that looked off.

I fed all 10 straight into `RegexParser`, skipping the `JavaCharacterParser` layer that normally unescapes Java string literals before the regex parser ever sees them, and diffed the result against real `java.util.regex.Pattern.compile(...)` on the same strings.

Every one of them matches:

- `[\]`, `[\]a`, `[\\\]`, `[\\\a`, `[\\\]a` come back as unclosed character classes - exactly what `Pattern` does too (it throws `Unclosed character class`).
- `[\]a]`, `[\\]`, `[\\\a]`, `[\\\]a]`, `[\\\\]` parse successfully, and the AST content matches what `Pattern` actually matches.
- The one concrete claim in the report - that `[\\\\]` "loses" a backslash while parsing - doesn't hold up. The AST has two separate escaped-backslash elements, not one.

Worth noting: the report itself admits it's only showing "the first element of the parsing." So whatever debug tool produced these examples was dumping a truncated AST, and that's almost certainly why some correct results looked broken.

I couldn't find an actual bug, so instead of touching parser logic I added one test per reported input, asserting both the resulting AST and the syntax errors. At minimum this locks down these specific edge cases so nobody has to re-litigate them.

## Test plan
- [x] `mvn -o test` in `regex-parsing` - passes, 9 new tests in `CharacterClassTreeTest`
- [x] `mvn -o test` from repo root - `BUILD SUCCESS`, no failures in any module

### DIFF
--- a/regex-parsing/src/test/java/org/sonarsource/analyzer/commons/regex/ast/CharacterClassTreeTest.java
+++ b/regex-parsing/src/test/java/org/sonarsource/analyzer/commons/regex/ast/CharacterClassTreeTest.java
@@ -140,6 +140,86 @@ class CharacterClassTreeTest {
     );
   }
 
+  // Regression tests for https://github.com/SonarSource/sonar-analyzer-commons/issues/212 (ACOMMONS-3):
+  // escaped ']' and chained backslashes inside a character class must behave exactly like java.util.regex.Pattern.
+
+  @Test
+  void unclosedClassStartingWithEscapedClosingBracket() {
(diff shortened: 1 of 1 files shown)


aurelien-coet-sonarsource (approved):
LGTM

PR #432, opened 2026-07-13
ACOMMONS-1 RuleMetadataLoader ignores fields other than "key" of annotation "org.sonar.check.Rule"

mallory-scotton:
## What

Fixes ACOMMONS-1: `RuleMetadataLoader` ignored every field of the `org.sonar.check.Rule` annotation except `key`.

## Why

`addRuleByAnnotatedClass` always read a JSON and an HTML file to fill in name, description, severity, tags and status, even when the `@Rule` annotation already carried those same values. That file read ran every time afterward and overwrote whatever the annotation had set. So a custom rule that relied only on the annotation, like the one in the ticket, just didn't work. Every rule needed a JSON+HTML pair no matter how much metadata was already sitting in the annotation.

## Fix

- When `RuleMetadataLoader` is built with the constructor that only takes a `SonarRuntime` (no resource folder), `addRulesByAnnotatedClass` skips JSON/HTML loading and keeps whatever `RulesDefinitionAnnotationLoader` already set from the annotation: name, description, priority/severity, tags, status.
- `addRulesByRuleKey` now throws a clear `IllegalStateException` when no resource folder is configured, since there's no annotation to fall back on in that path.
- Behavior with a resource folder is unchanged, existing consumers aren't affected.

Rule type isn't part of the `@Rule` annotation, but the SonarQube API already infers it from tags ("bug" becomes BUG, "security" becomes VULNERABILITY) or falls back to CODE_SMELL. That's enough to cover the ticket's example without adding anything on our side.

## Testing

- Test mirroring the ticket's example: name/description/priority/tags coming only from the annotation.
- Test for the default fallbacks (MAJOR severity, READY status, CODE_SMELL type) when the annotation doesn't set them.
- Test for the new error when `addRulesByRuleKey` is called without a resource folder.
- Updated the module README to mention the annotation-only mode.

### DIFF
--- a/commons/README.md
+++ b/commons/README.md
@@ -2,7 +2,7 @@ SonarSource Analyzers Commons (compatible with SQ >=7.9)
 =========================
 Logic useful for an average language plugin
 
-* [`RuleMetadataLoader`](./src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java) - to define rules metadata based on `json` and `html` files
+* [`RuleMetadataLoader`](./src/main/java/org/sonarsource/analyzer/commons/RuleMetadataLoader.java) - to define rules metadata based on `json` and `html` files, or, when constructed without a resource folder, directly from the `org.sonar.check.Rule` annotation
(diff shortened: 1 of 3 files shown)


aurelien-coet-sonarsource (approved):
LGTM

Wohops:
Thank you for fixing this, guys! It was waiting in our backlog since 5 years! :)

PR #433, opened 2026-07-13
ACOMMONS-4 Preserve whitespace inside character classes in free-spacing mode

mallory-scotton:
## Summary

Fixes ACOMMONS-4. Enabling the comment/extended regex flag (`x` / `Pattern.COMMENTS`) made `RegexLexer` strip whitespace and `#` comments from inside character classes (`[...]`), not just outside them.

Under PCRE-style `/x` semantics, which this library uses to model PHP, Python, and Ruby regexes, whitespace inside a character class is a literal character, not something to ignore. The lexer removed it anyway, so `[a ]` and `[a]` produced the same AST. That made `SingleCharCharacterClassFinder` flag `[a ]` as redundant, when collapsing it to `a` would actually drop a space that was part of the pattern.

I checked this against real PCRE and Python's `re.VERBOSE`. Both preserve whitespace in character classes under the extended flag, so this was a genuine bug in the lexer, not just a blind spot in this one finder. Fixed it in `RegexParser.parseCharacterClass()`, which now turns off free-spacing mode while parsing between `[` and `]` and turns it back on afterward. Same save/restore pattern already used for inline `(?x:...)` flag groups.

**Update:** an AI-assisted review caught a follow-up bug in the first version of this fix: free-spacing mode was disabled *after* the opening `[` had already been consumed, so the lexer's lookahead skipped whitespace immediately after `[` before the mode-suspend took effect. `[ a]` still collapsed to `[a]`, and `[ ^a]` was wrongly treated as negated. Fixed by disabling free-spacing mode before `[` is read instead of after, so the buffer reset rewinds far enough to preserve that leading whitespace.

## Changes

- `RegexParser.parseCharacterClass()`: suspend free-spacing mode before reading the opening `[`, restore it after the closing `]`.
- `CharacterClassTreeTest`: regression tests for whitespace preservation, `#` no longer being read as a comment inside a class, free-spacing mode resuming correctly afterward, and leading whitespace (including whitespace before `^` no longer being mistaken for negation).
- `SingleCharCharacterClassFinder.yml`: fixture cases reproducing the ticket's false positive (`[a ]` and `[ a]` under `x` are compliant, `[a]` under `x` is still flagged).

## Test plan

- [x] New unit tests in `CharacterClassTreeTest` for whitespace/comment handling under `Pattern.COMMENTS`, including leading whitespace and whitespace before `^`
- [x] New fixture cases in `SingleCharCharacterClassFinder.yml`
- [x] Reverted the parser change locally to confirm the new tests fail without it, then confirmed they pass with it
- [x] Full `regex-parsing` module test suite passes, no regressions

### DIFF
--- a/regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/RegexParser.java
+++ b/regex-parsing/src/main/java/org/sonarsource/analyzer/commons/regex/RegexParser.java
@@ -933,11 +933,11 @@ public class RegexParser {
   }
 
   protected CharacterClassTree parseCharacterClass() {
-    SourceCharacter openingBracket = characters.getCurrent();
-    characters.moveNext();
     // Whitespace and comments are not stripped inside character classes, even in free-spacing mode.
     boolean previousFreeSpacingMode = characters.getFreeSpacingMode();
     characters.setFreeSpacingMode(false);
+    SourceCharacter openingBracket = characters.getCurrent();
+    characters.moveNext();
     boolean negated = false;
     if (characters.currentIs('^')) {
       characters.moveNext();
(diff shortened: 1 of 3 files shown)


aurelien-coet-sonarsource (approved):
LGTM

PR #437, opened 2026-07-15
ACOMMONS-84 Honor sonar.secrets.disableTestFileDetection in TestFileClassifier

teemu-rytilahti-sonarsource:
Makes the shared `TestFileClassifier` in analyzer-commons honor the sonar-text secrets analyzer property `sonar.secrets.disableTestFileDetection`, in addition to the existing generic `sonar.testFileHeuristic.disabled` opt-out. Either key now disables the path-based test-file heuristic, so every analyzer that uses `TestFileClassifier` (go, iac, xml, tsql, slang, …) gets the behavior — the gating stays in commons rather than being reimplemented per analyzer.

Adds the public constant `HEURISTIC_DISABLED_KEY_SONARTEXT`, OR's it into `isTestSourceConfigured(...)` alongside the existing key, and adds a test mirroring the existing opt-out test.

Note that `sonar.secrets.*` is a secrets-scoped namespace but `TestFileClassifier` is shared beyond secrets analysis; this intentionally lets the property disable the heuristic for all consumers.


teemu-rytilahti-sonarsource:
This is in incorrect place, instead of honoring sonar texts setting for test files, sonar-text should follow ours.

PR #441, opened 2026-07-17
CLP-212: Migrate scanner test fixtures

vdiez:
## Summary

- replace the SonarQube implementation artifact with scanner-engine 13.4.1 fixtures
- migrate moved scanner implementation imports and use `TestSonarRuntime`
- update the XML test verifier module so consumers can use the new fixture types

## Dependency changes

Before: `commons`, `xml-parsing`, and `test-xml-parsing` depended on `org.sonarsource.sonarqube:sonar-plugin-api-impl`.

After: the modules directly declare `plugin-api-scanner-impl` and `sensor-test-fixtures` at `13.4.1.4007`. `test-xml-parsing` intentionally exposes these at compile scope because its published verifier API constructs and returns scanner fixture types. The resolved Maven graph contains no `sonar-plugin-api-impl`.

`sonar-plugin-api-test-fixtures` remains separate and is retained only where its logging helpers are used.

## Validation

- affected `commons`, `xml-parsing`, and `test-xml-parsing` reactor tests
- Maven dependency-tree check for `org.sonarsource.sonarqube:sonar-plugin-api-impl` (no matches)
- `git diff --check`

### DIFF
--- a/commons/pom.xml
+++ b/commons/pom.xml
@@ -44,8 +44,13 @@
 
     <!-- TEST -->
     <dependency>
-      <groupId>org.sonarsource.sonarqube</groupId>
-      <artifactId>sonar-plugin-api-impl</artifactId>
+      <groupId>org.sonarsource.scanner.engine</groupId>
+      <artifactId>plugin-api-scanner-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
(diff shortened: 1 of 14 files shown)


aurelien-coet-sonarsource (approved):
LGTM

PR #458, opened 2026-07-21
ACOMMONS-101 Add MuleSoft schema host to cleartext namespace URI authorities

teemu-rytilahti-sonarsource:
Adds `www.mulesoft.org` to the `NAMESPACE_URI_AUTHORITIES` allowlist in `CleartextProtocolFilter`.

MuleSoft Mule config files declare XML namespaces like `http://www.mulesoft.org/schema/mule/core`. These are opaque namespace identifiers, not real cleartext connections, so they should not trigger a cleartext-protocol warning.


daniel-teuchert-sonarsource (approved):
LGTM! Thanks!

PR #472, opened 2026-07-23
ACOMMONS-110 Remove duplicate regex from Secret Classifier

daniel-teuchert-sonarsource:
This PR removes an unnecessary regex from the Fake Value group of the Secret Classifier.
Since we have a check for min length 5, the regular expression detecting a secret that is only consisting of one character is not matching any cases that are not also detected by `"(?<char>[\\w\\*\\.])\\k<char>{3}"`.


teemu-rytilahti-sonarsource (approved):
LGTM! 👍

PR #482, opened 2026-08-11
SONARIAC-3061 Treat single-label hostnames as safe in CleartextProtocolFilter

daniel-teuchert-sonarsource:
## Summary
- S5332 raised false positives on single-label HTTP hostnames (e.g. `http://estimate-translations-dev/meanings`) — Kubernetes/Docker Compose service names that cannot resolve on the public internet.
- `CleartextProtocolFilter.isSafeWithoutTls()` now treats single-label hostnames (no dot) as safe, generalising the existing `localhost` special case. Excludes IPv6 literals (`[...]`) and numeric IP literals — decimal (e.g. `2130706433`) and hexadecimal (e.g. `0x7f000001`) — which are IP addresses rather than DNS names. Octal single-integer literals are already covered since they're all decimal digits.
- Scoped to hosts parsed strictly via `java.net.URI#getHost()` only (kept out of the shared `isSafeHost` helper). The lenient string-fallback path (used for template placeholders / underscore hostnames) is untouched, since it can extract a truncated fake host from malformed input (e.g. `http://foo bar.com` → `foo`) that must not be trusted by this rule.

Jira: SONARIAC-3061

## Test plan
- [x] Added parameterized cases for single-label hostnames to `safeUrls`/`safeUris`
- [x] Added exclusion cases (decimal-IP, hexadecimal-IP, public IPv6 literal) to `notSafeUrls`/`notSafeUris`
- [x] `mvn -pl commons test -Dtest=CleartextProtocolFilterTest` passes

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
@@ -277,15 +277,15 @@ public final class CleartextProtocolFilter {
     if (host == null) {
       return false;
     }
-    return isSafeHost(host) || isSingleLabelHost(host);
+    return isSafeHost(host);
   }
 
   private static boolean isSafeHost(String host) {
-    return isInternalHost(host) || isNamespaceUriAuthority(host) || isDocumentationHost(host);
+    return isInternalHost(host) || isNamespaceUriAuthority(host) || isDocumentationHost(host) || isSingleLabelHost(host);
   }
 
   // Single-integer IPv4 literal written in hexadecimal, e.g. 0x7f000001 (== 127.0.0.1).
-  private static final Pattern HEX_IP_LITERAL = Pattern.compile("^0[xX][0-9a-fA-F]+$");
+  private static final Pattern HEX_IP_LITERAL = Pattern.compile("^0[xX][0-9a-fA-F]++$");
 
   // Single-label hostnames (no dot) cannot resolve on the public internet; excludes IPv6
   // literals ("[...]") and numeric IP literals (decimal, e.g. 2130706433, and hexadecimal,
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
@@ -284,17 +284,22 @@ public final class CleartextProtocolFilter {
     return isInternalHost(host) || isNamespaceUriAuthority(host) || isDocumentationHost(host);
   }
 
+  // Single-integer IPv4 literal written in hexadecimal, e.g. 0x7f000001 (== 127.0.0.1).
+  private static final Pattern HEX_IP_LITERAL = Pattern.compile("^0[xX][0-9a-fA-F]+$");
+
   // Single-label hostnames (no dot) cannot resolve on the public internet; excludes IPv6
-  // literals ("[...]") and all-digit hosts, which are IP addresses rather than DNS names.
+  // literals ("[...]") and numeric IP literals (decimal, e.g. 2130706433, and hexadecimal,
+  // e.g. 0x7f000001 — octal single-integer literals are all-digit and match the decimal case
+  // too), which are IP addresses rather than DNS names.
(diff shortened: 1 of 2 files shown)


commons/src/test/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilterTest.java
```diff
@@ -276,7 +283,12 @@ static Stream<URI> notSafeUris() {
       URI.create("http://localhost@evil.com"),
       URI.create("http://www.w3.org@evil.com"),
       // Opaque URI — http scheme but no host component (getHost() == null)
-      URI.create("http:not-hierarchical")
+      URI.create("http:not-hierarchical"),
+      // Single-label host exclusions — IP addresses, not DNS names
+      URI.create("http://2130706433"),
+      URI.create("http://[2001:db8::1]"),
+      URI.create("http://0x7f000001"),
```
teemu-rytilahti-sonarsource: This is localhost, should it be safe albeit esoteric? I agree that we can ignore this, but it would be nice to document it explicitly.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
```diff
@@ -274,13 +277,32 @@ public static boolean isSafeWithoutTls(URI url) {
     if (host == null) {
       return false;
     }
-    return isSafeHost(host);
+    // isSingleLabelHost is deliberately called here rather than folded into isSafeHost: isSafeHost
+    // is also used by the lenient string-fallback path in isSafeWithoutTls(String), which can
+    // extract a truncated fake "host" from malformed input (e.g. a URL like http://user:pass@${placeholder})
+    // that must not be trusted by this rule.
```
teemu-rytilahti-sonarsource: Feels overly expressive, and even after reading I am not much wiser without looking at isSafeWithoutTls separately.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
```diff
@@ -274,13 +277,32 @@ public static boolean isSafeWithoutTls(URI url) {
...
+    return isSafeHost(host) || isSingleLabelHost(host);
   }
 
   private static boolean isSafeHost(String host) {
     return isInternalHost(host) || isNamespaceUriAuthority(host) || isDocumentationHost(host);
   }
 
+  // Single-integer IPv4 literal written in hexadecimal, e.g. 0x7f000001 (== 127.0.0.1).
+  private static final Pattern HEX_IP_LITERAL = Pattern.compile("^0[xX][0-9a-fA-F]++$");
```
teemu-rytilahti-sonarsource, line 290: Should we have a max-length limit here?

teemu-rytilahti-sonarsource (approved):
LGTM, nice improvement! Added a couple of nits, feel free to act as you wish.

commons/src/test/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilterTest.java
daniel-teuchert-sonarsource: Good point. We can probably also ignore this variant of localhost and I can give different hex example here.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: I intentionally placed a comment here as I thought it might be confusing to why we not just add the isSingleLableHost to the checks in isSafeHost.
I can also remove it.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: I don't think so as this pattern should only run on already extracted URLs not full files.

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: Removed.

PR #494, opened 2026-08-21
ACOMMONS-131 Recognize kebab-case cloud resource identifiers as non-secrets

daniel-teuchert-sonarsource:
## Why

User feedback triage surfaced a false positive on a hardcoded-secrets rule (S6418, generic YAML): a value shaped like an Azure resource-naming-convention identifier — hyphen-segmented, ending in a short region abbreviation plus a zero-padded instance number (e.g. `src-test-srv-dev-uks-001`) — was flagged because the sibling key contained a credential-related word (e.g. "...Credential").

It slipped through `SecretClassifier.isKnownNonSecret()` because the existing `STRUCTURED_FORMAT` path-pattern only recognizes `/`-separated paths, not `-`-separated identifiers, and the abbreviated, non-dictionary hyphen segments (e.g. 3-4 letter region/component codes) score as "random enough" / "not human language" just like a real secret would.

## What

Added a new pattern to the `STRUCTURED_FORMAT` category in `SecretClassifier`:

```
^[a-z][a-z0-9]*+(?:-[a-z0-9]++){2,}-[a-z]{2,4}-\d{2,4}$
```

This requires at least 5 lowercase alphanumeric hyphen-separated segments, with the value ending in a short (2-4 char) alpha segment followed by a 2-4 digit numeric suffix. This is the structural shape shared by cloud resource-naming conventions across providers — it intentionally does **not** hardcode any specific region/provider vocabulary (e.g. no list of Azure region abbreviations), since that would be unmaintainable and provider-specific.

The pattern is conservative by construction: a real high-entropy secret (base64, hex, random alphanumeric, UUID) essentially never has this many hyphen-separated segments capped by a short alpha suffix and a zero-padded numeric instance id, so the risk of suppressing genuine secrets is low.

This is `SecretClassifier`, shared across every analyzer calling `isKnownNonSecret()`, so the change was kept as narrow as possible and validated against the existing test corpus.

## Test evidence

- Added negative cases to `shouldNotClassifyRealisticTokensAsNonSecrets` proving the new pattern does **not** catch real-looking secrets: a 32-char hex token, a base64 token, and a UUID-shaped value.
- Ran the full `commons` test suite (not just the new test) — all tests pass, no existing "should still be flagged" case flipped to non-secret.

```
mvn -pl commons -am test
```
passes with no failures.

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/SecretClassifier.java
@@ -187,7 +187,7 @@ public final class SecretClassifier {
       // Stopgap: such values may instead be excluded by ignoring lockfiles by path.
       "^v?\\d++(?:\\.\\d++)++(?:\\([^()]*+\\))++$",
       // Kebab-case cloud-resource identifiers following a naming convention, ending in a short region/environment-style
-      // abbreviation plus a zero-padded instance number, e.g. "srv-creds-dbac-dsp-dev-uks-001". At least 5 lowercase
+      // abbreviation plus a zero-padded instance number. At least 5 lowercase
       // alphanumeric segments are required so the shape stays specific: a high-entropy secret essentially never has
       // this many hyphen-separated segments capped by a short alpha suffix and a padded numeric instance id.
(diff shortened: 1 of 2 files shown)


daniel-teuchert-sonarsource:
This is too broad and could lead to FNs.

PR #496, opened 2026-08-25
ACOMMONS-139 CleartextProtocolFilter: recognise the Android emulator network as safe for cleartext protocols

teemu-rytilahti-sonarsource:
Recognise the Android emulator's special network as a safe host in `CleartextProtocolFilter.isSafeWithoutTls`, alongside the existing loopback and cloud-IMDS internal-host cases.

## Details
- `SAFE_HOSTS` matches only the specific addresses the Android emulator networking docs define (router, host-loopback alias, DNS, device), not the whole `10.0.2.0/24`, so real internal hosts elsewhere in that subnet aren't silently suppressed.
- Class Javadoc's "Internal hosts" bullet mentions the Android emulator network; the intro no longer hardcodes a category count.
- Fixed `isSafeWithoutTls(String)`'s lenient fallback path to also apply the single-label-hostname exemption (previously only the `URI` overload did), guarded so a malformed/truncated host from that fallback can't be mistaken for a real single-label host.
- Added tests for the narrowed emulator-address matching, the fallback-path fix, and their edge cases.

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
@@ -37,7 +37,7 @@ import java.util.stream.Collectors;
  * {@code stomp}. Any other scheme is considered safe (e.g. {@code https}, {@code wss},
  * {@code sftp}).
  * <p>
- * Three categories of safe cleartext URLs are recognised:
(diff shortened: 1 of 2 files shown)


daniel-teuchert-sonarsource (approved):

daniel-teuchert-sonarsource (approved):
LGTM!
Not sure if the adjustments to the handling of unparsable URLs is worth it.

PR #511, opened 2026-09-07
ACOMMONS-153 Add jabber.org and apache.org XML feature URLs as safe cleartext identifiers

daniel-teuchert-sonarsource:
Part of 

## What
`CleartextProtocolFilter` now recognises a new category of well-known identifier URLs: specific `http://` URLs (or URL prefixes) that are opaque protocol/feature identifiers rather than real network endpoints, on hosts that also serve ordinary content:

- `http://jabber.org/protocol/*` — XMPP protocol namespaces (XEPs)
- `http://apache.org/xml/features/nonvalidating/load-dtd-grammar`
- `http://apache.org/xml/features/nonvalidating/load-external-dtd`
  (Xerces/JAXP XML parser features to disable DTD loading)

## Why
These are used as opaque literal identifiers (XML namespace URIs / XML parser feature-flag URIs) in source code, never as actual network requests, so flagging them as insecure cleartext protocol usage is a false positive.

## How
Unlike the existing `NAMESPACE_URI_AUTHORITIES` list (matched on host only), `jabber.org` and `apache.org` also serve ordinary content outside these specific paths, so a new `SAFE_URL_PATH_PREFIXES` pattern matches on host + path instead. Added positive and boundary-negative test cases (bare host, near-miss suffix, lookalike host) to `CleartextProtocolFilterTest`.

## Testing
`mvn -pl commons -am test -Dtest=CleartextProtocolFilterTest` → 214 tests, 0 failures.

### DIFF
--- a/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
+++ b/commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
@@ -164,14 +164,15 @@ public final class CleartextProtocolFilter {
 
   // --- Well-known identifier URLs (host + path) -------------------------------------------
   // Unlike the namespace URI authorities above, these hosts also serve ordinary content,
-  // so only the specific path(s) below — opaque protocol/feature identifiers, not real
-  // endpoints — are considered safe. Matched against "host + path" of the URI.
+  // so only the specific path prefix(es) below — opaque protocol/feature identifiers, not
+  // real endpoints — are considered safe. Matched against "host + path" of the URI.
   private static final Pattern SAFE_URL_PATH_PREFIXES = Pattern.compile("(?:" +
     // XMPP protocol namespaces (XEPs)
     "^jabber\\.org/protocol/|" +
-    // Xerces/JAXP XML parser features to disable DTD loading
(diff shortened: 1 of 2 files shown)


commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
```diff
@@ -282,7 +303,10 @@ public static boolean isSafeWithoutTls(URI url) {
     if (host == null) {
       return false;
     }
-    return isSafeHost(host) || isSingleLabelHost(host);
+    if (isSafeHost(host) || isSingleLabelHost(host)) {
+      return true;
+    }
+    return isKnownIdentifierUrl(scheme + "://" + host + url.getRawPath());
```
teemu-rytilahti-sonarsource: Is this necessary after the latest changes? Your call, looks much cleaner than earlier!

teemu-rytilahti-sonarsource (approved):

commons/src/main/java/org/sonarsource/analyzer/commons/appsec/CleartextProtocolFilter.java
daniel-teuchert-sonarsource: Not at all. Thanks for pointing this out.

